PR #3, opened 2022-09-02
DATAGO-38922: "Flatten" the Solace scan results so they are just configuration and no wrapper

gregmeldrum:
The Solace scan results had an extra wrapper layer which would make the processing in Event Portal more difficult.

Removed the extra layer around the Solace configuration data

 REST API

Also fixed up the base docker build file and removed the JavaRTO test since it does not always work on mac M1 devices.


service/application/docker/base-image/buildBaseImage.sh
```diff
@@ -5,5 +5,4 @@ if [ -z "$1" ]
     echo "usage: buildBaseImage <tag>"
     exit 1
 fi
-docker build . -t 868978040651.dkr.ecr.us-east-1.amazonaws.com/runtime-agent-base:$1
-#docker push 868978040651.dkr.ecr.us-east-1.amazonaws.com/runtime-agent-base:$1
\ No newline at end of file
+docker build . -t runtime-agent-base:$1
```
gregmeldrum, line 8: Remove the amazon URL since this is now open-source.

service/application/src/test/java/com/solace/maas/ep/runtime/agent/messagingServices/RtoMessagingServiceTests.java
```diff
@@ -34,7 +35,7 @@ public class RtoMessagingServiceTests {
     @Mock
     SessionEventCallback sessionEventCallback;
 
-
+    @Ignore
```
gregmeldrum, line 38: Some M1 macs cannot execute this test so @Ignore since JavaRTO is not currently supported.

PR #8, opened 2022-09-07
Add Swagger configuration and docs

AHabes:
Add Swagger documentation for the runtime agent REST APIs.

Added Swagger config and other related descriptions and files.

Manual

N/A

### DIFF
--- a/service/application/src/main/resources/application.yml
+++ b/service/application/src/main/resources/application.yml
@@ -1,3 +1,10 @@
+springdoc:
+  packages-to-scan: com.solace.maas.ep.runtime.agent.scanManager
+  api-docs:
+    path: /docs/runtime-agent
+  swagger-ui:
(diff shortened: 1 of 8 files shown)


service/application/src/main/resources/application.yml
```diff
@@ -1,3 +1,10 @@
+springdoc:
+  packages-to-scan: com.solace.maas.ep.runtime.agent.scanManager
```
slunelsolace, line 2: Wouldn't this be "com.solace.maas.ep.runtime.agent.scanManager.rest"?

AHabes: Thanks! `packages-to-scan` has been updated.

slunelsolace (approved):
Going to approve. Seems like Sonar is acting up.

PR #12, opened 2022-09-13
Send scan data to the Event Portal via MOP messages.

AHabes:
Send the scan data to EP via MOP messages.

Add the Scan data message processor, publisher, and route builder.

Manual + Unit tests.

N/A

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/runtime/agent/processor/ScanDataProcessor.java
+++ b/service/application/src/main/java/com/solace/maas/ep/runtime/agent/processor/ScanDataProcessor.java
@@ -50,8 +50,8 @@ public class ScanDataProcessor implements Processor {
         topicDetails.put("orgId", orgId);
         topicDetails.put("runtimeAgentId", runtimeAgentId);
         topicDetails.put("messagingServiceId", messagingServiceId);
-        topicDetails.put(" scanId", scanId);
-        topicDetails.put(" dataCollectionType", dataCollectionType);
+        topicDetails.put("scanId", scanId);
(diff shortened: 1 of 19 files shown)


gregmeldrum (approved):
Looks good!

service/application/src/main/java/com/solace/maas/ep/runtime/agent/processor/ScanDataProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+        String scanId = (String) properties.get(RouteConstants.SCAN_ID);
+        String dataCollectionType = (String) properties.get(RouteConstants.SCAN_TYPE);
+
+        ScanDataMessage scanDataMessage =
+                new ScanDataMessage(orgId, scanId, dataCollectionType, body, Instant.now().toString());
+
+        topicDetails.put("orgId", orgId);
+        topicDetails.put("runtimeAgentId", runtimeAgentId);
+        topicDetails.put("messagingServiceId", messagingServiceId);
+        topicDetails.put(" scanId", scanId);
```
gregmeldrum, line 53: Is the space intentional? I see it on the put and get.

gregmeldrum (approved):
Looks good. Just one question about the encoding/decoding of the topic details.

service/application/src/main/java/com/solace/maas/ep/runtime/agent/processor/ScanDataProcessor.java
AHabes: Thanks for the catch, the space has been removed.

PR #16, opened 2022-09-20
The Event Management Agent should be case in-sensitive

OmarMostafa125:
When the configuration file is generated from the runtime agent we specify the messaging service type as solace or kafka "lowercase". The runtime should accept the types as case in sensitive


AHabes (approved):
E2E test looks good.

PR #19, opened 2022-09-26
Send the scan logs to EP

AHabes:
What is the purpose of this change?
Send the scan logs to EP via MOP messages.

How was this change implemented?
Add the Logging appender, the logs data message processor, publisher, and route builder.

How was this change tested?
Manual + Unit tests.

Is there anything the reviewers should focus on/be aware of?
N/A

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/common/messages/ScanLogsMessage.java
+++ b/service/application/src/main/java/com/solace/maas/ep/common/messages/ScanLogsMessage.java
@@ -1,0 +1,29 @@
+package com.solace.maas.ep.common.messages;
+
+import com.solace.maas.ep.runtime.agent.plugin.mop.MOPMessage;
+import com.solace.maas.ep.runtime.agent.plugin.mop.MOPMessageType;
+import com.solace.maas.ep.runtime.agent.plugin.mop.MOPProtocol;
+import com.solace.maas.ep.runtime.agent.plugin.mop.MOPUHFlag;
+
+public class ScanLogsMessage extends MOPMessage {
+    String orgId;
+
+    String scanId;
+
+    String log;
+
+    String timestamp;
+
+    public ScanLogsMessage(String orgId, String scanId, String log, String timestamp) {
+        super();
+        withMessageType(MOPMessageType.generic)
+                .withProtocol(MOPProtocol.event)
+                .withVersion("1")
+                .withUhFlag(MOPUHFlag.ignore);
+
+        this.orgId = orgId;
+        this.scanId = scanId;
+        this.log = log;
+        this.timestamp = timestamp;
+    }
+}
--- a/service/application/src/main/java/com/solace/maas/ep/runtime/agent/processor/ScanLogsProcessor.java
+++ b/service/application/src/main/java/com/solace/maas/ep/runtime/agent/processor/ScanLogsProcessor.java
@@ -1,0 +1,64 @@
+package com.solace.maas.ep.runtime.agent.processor;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import com.solace.maas.ep.common.messages.ScanLogsMessage;
+import com.solace.maas.ep.runtime.agent.config.eventPortal.EventPortalProperties;
+import com.solace.maas.ep.runtime.agent.plugin.constants.RouteConstants;
+import com.solace.maas.ep.runtime.agent.publisher.ScanLogsPublisher;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.camel.Exchange;
+import org.apache.camel.Processor;
+import org.json.JSONObject;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+
+@Slf4j
+@Component
+public class ScanLogsProcessor implements Processor {
+    private final String orgId;
+    private final String runtimeAgentId;
+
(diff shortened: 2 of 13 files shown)


service/application/src/main/java/com/solace/maas/ep/runtime/agent/service/ScanService.java
```diff
@@ -136,10 +146,14 @@ public String singleScan(List<RouteBundle> routeBundles, int numExpectedCompleti
...
+                Map<String, String> details = new HashMap<>();
                 MDC.put(SCAN_ID, scanId);
-                fileLoggerFactory.create();
-                LoggingProcessor loggingProcessor = new LoggingProcessor();
-                loggingService.addLoggingProcessor(scanId, loggingProcessor);
+                details.put(SCAN_ID, scanId);
+                details.put(MESSAGING_SERVICE_ID, routeBundle.getMessagingServiceId());
+                details.put(SCHEDULE_ID, groupId);
+
+                createScanFileLogger(details);
```
slunelsolace, line 155: Why is this still being done here? What happens if we use another method to perform a scan that isn't a single scan. Do we not get logs any more?

slunelsolace (dismissed):
StreamingAppender hasn't been added to the logback configuration file. This should be done instead of injecting the loggers into the "singleScan" method.

service/application/src/main/java/com/solace/maas/ep/runtime/agent/processor/ScanLogsProcessor.java
```diff
@@ -0,0 +1,65 @@
...
+                new ScanLogsMessage(orgId, scanId, json.toString(), timestamp);
+
+        topicDetails.put("orgId", orgId);
+        topicDetails.put("runtimeAgentId", runtimeAgentId);
+        topicDetails.put("messagingServiceId", messagingServiceId);
+        topicDetails.put("scanId", scanId);
+
+        logDataPublisher.sendScanLogData(logDataMessage, topicDetails);
+
+        log.info("Scan logs have been published successfully.");
```
moodiRealist, line 63: Would this create a log message for every log we publish? It might be many logs at Info level, even at debug level I think the log needs some info to be useful for debugging purposes

service/application/src/main/java/com/solace/maas/ep/common/messages/ScanLogsMessage.java
```diff
@@ -0,0 +1,29 @@
...
+import com.solace.maas.ep.runtime.agent.plugin.mop.MOPUHFlag;
+
+public class ScanLogsMessage extends MOPMessage {
+    String orgId;
+
+    String scanId;
+
+    String log;
+
+    String timestamp;
```
moodiRealist, line 15: What is the format of this timestamp? Is it in Epoch Milliseconds? If yes, can we change the type to `Long` instead?

slunelsolace:
@AHabes Hey we should talk about setting MDC values across threads. The logger is losing log messages between SEDA endpoints. This is easy enough to fix.

service/application/src/main/java/com/solace/maas/ep/common/messages/ScanLogsMessage.java
AHabes: The timestamp key that was part of the json object was of type "Instant" converted to a String. However, after removing the json object, the timestamp is now parsed from the log event, which is of type long.

service/application/src/main/java/com/solace/maas/ep/runtime/agent/scanManager/ScanManager.java
```diff
@@ -65,7 +70,9 @@ public String scan(ScanRequestBO scanRequestBO) {
                         .stream())
                 .collect(Collectors.toUnmodifiableList());
 
-        return scanService.singleScan(routes, routes.size(), scanId);
+        loggingService.prepareLoggers(groupId, scanId, messagingServiceEntity.getId());
```
slunelsolace, line 73: This hasn't addressed my previous point about not creating the Logger for each scan in specific methods. We need to use the object created by Logback and not create a Logger for each scan initiated.

service/plugin/src/main/java/com/solace/maas/ep/runtime/agent/plugin/processor/logging/MDCProcessor.java
```diff
@@ -0,0 +1,22 @@
...
+import com.solace.maas.ep.runtime.agent.plugin.constants.RouteConstants;
+import org.apache.camel.Exchange;
+import org.apache.camel.Processor;
+import org.slf4j.MDC;
+import org.springframework.stereotype.Component;
+
+@Component
+public class MDCProcessor implements Processor {
+    @Override
+    public void process(Exchange exchange) throws Exception {
```
slunelsolace, line 12: Isn't ScanType missing here?

service/application/src/main/java/com/solace/maas/ep/runtime/agent/service/ScanService.java
AHabes: The coe

slunelsolace (approved):

service/plugin/src/main/java/com/solace/maas/ep/runtime/agent/plugin/processor/logging/MDCProcessor.java
AHabes: @slunelsolace, Thanks for the follow up, I have updated the interceptor to set the scanType in the header and updated the MDCProcessor accordingly.

PR #23, opened 2022-10-17
Implement route completion and scan status logic

AHabes:
**What is the purpose of this change?**
Implement the route completion and scan status logic. Also, send the scan status to EP.

**How was this change implemented?**
Update the route completion logic, implement `ScanDataStatusMessage`, `ScanStatusProcessor`, and `ScanStatusPublisher`. Also, update the Camel routes.

**How was this change tested?**
Manual + unit tests.

**Is there anything the reviewers should focus on/be aware of?**
N/A


service/application/src/main/java/com/solace/maas/ep/event/management/agent/service/ScanService.java
```diff
@@ -114,11 +115,17 @@ public ScanService(ScanRepository repository, ScanRouteService scanRouteService,
...
      */
-    public String singleScan(List<RouteBundle> routeBundles, int numExpectedCompletionMessages, String groupId, String scanId) {
+    public String singleScan(List<RouteBundle> routeBundles, String groupId, String scanId) {
         log.info("Starting single scan request {}.", scanId);
 
         ScanEntity savedScanEntity = null;
 
+        List<String> scanTypes = parseScanTypes(routeBundles, new ArrayList<>());
+
+        routeCompleteProcessor.sendScanStatus(scanId, scanTypes, groupId,
```
slunelsolace, line 125: Why call a Camel Processor directly?

service/kafka-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/kafka/route/handler/KafkaClusterConfigurationDataPublisherRouteBuilder.java
```diff
@@ -13,8 +14,9 @@
 public class KafkaClusterConfigurationDataPublisherRouteBuilder extends DataPublisherRouteBuilder {
     @Autowired
     public KafkaClusterConfigurationDataPublisherRouteBuilder(KafkaClusterConfigurationProcessor processor,
-                                                              RouteManager routeManager, MDCProcessor mdcProcessor) {
+                                                              RouteManager routeManager, MDCProcessor mdcProcessor,
+                                                              RouteCompleteProcessor routeCompleteProcessor) {
         super(processor, KafkaRouteId.KAFKA_CLUSTER_CONFIGURATION.label, KafkaRouteType.KAFKA_CUSTER_CONFIGURATION.label,
-                routeManager, mdcProcessor);
+                routeManager, mdcProcessor, routeCompleteProcessor);
```
slunelsolace, line 20: Doing this to all of these endpoints isn't really necessary. If you create a single RouteBuilder that handles all "Scan Status" messages you can avoid injecting your routeCompleteProcessor into every single plugin.

Take a look at the "DataCollectionFileWriteRouteBuilder" class.

service/plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/route/handler/base/DataAggregationRouteBuilder.java
```diff
@@ -100,7 +106,18 @@ public void configure() {
...
+                .shareUnitOfWork()
+                .to("seda:endOf" + routeId);
+
+        from("seda:endOf" + routeId + "?blockWhenFull=true&size=100")
+                .choice().when(header("DATA_PROCESSING_COMPLETE").isEqualTo(true))
+                .to("seda:processEndOf" + routeId)
+                .endChoice()
+                .end();
+
+        from("seda:processEndOf" + routeId + "?blockWhenFull=true&size=100")
```
slunelsolace, line 118: You don't need to use the "routeId" in this seda endpoint. Sending Scan Statuses does not need to be a dynamic endpoint generated at run time. Refer to the "DataCollectionFileWriteRouteBuilder". It does not need a dynamic route name for every scan type.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/RouteCompleteProcessorImpl.java
```diff
@@ -32,13 +40,69 @@ public void process(Exchange exchange) throws Exception {
...
+
+            save(scanStatusEntity);
+
+            sendScanDataStatus(scanId, scanType, groupId, messagingServiceId, ScanStatus.COMPLETE, ScanStatusType.PER_ROUTE);
+            log.info("Route {} completed for scan request {}", scanType, scanId);
         }
     }
+
+    private RouteEntity createScanStatusRoute(String scanType, String messagingServiceId) {
+        MessagingServiceRouteDelegate delegate = PluginLoader.findPlugin("SCAN_STATUS");
```
slunelsolace, line 62: Why is this a plugin? Sending Scan Statuses hasn't been created as a separate plugin module. You either need to create a separate plugin module and make it a plugin, or just leave the code inside of the EMA (Application Module) and not have it as a plugin.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/plugin/ScanStatusPublisherDelegateImpl.java
```diff
@@ -0,0 +1,21 @@
...
+@Component
+public class ScanStatusPublisherDelegateImpl extends MessagingServiceRouteDelegateImpl {
+    public ScanStatusPublisherDelegateImpl() {
+        super("SCAN_STATUS");
+    }
+
+    @Override
+    public List<RouteBundle> generateRouteList(List<RouteBundle> destinations, List<RouteBundle> recipients,
+                                               String scanType, String messagingServiceId) {
+        return List.of(createRouteBundle(destinations, recipients, scanType, messagingServiceId,
```
slunelsolace, line 18: You've made sending scan status a plugin. This kind of doesn't makes sense and I think your RouteBundle builder reflects this. There's no point in this delegate class as there's no enumeration value to examine.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/service/ScanService.java
slunelsolace: Also, does this work properly when set to standalone? I noticed your processors don't get instantiated if it is.

slunelsolace (changes requested):
Can you remove the plugin logic for Scan Status? It really doesn't seem like it needs to be a plugin. I would look at how DataCollectionFileWriteRouteBuilder writes its files. Scan Status can be handled in a similar manner but you don't need to make it a plugin. It could just become part of the EMA code.

The other Route Builders become much simpler then. They just need to send to your Scan Status Route and you don't have to dynamically create endpoints by appending the routeId to them.

slunelsolace (approved):

PR #30, opened 2022-12-01
Enhance the scan status and error reporting from EMA

AHabes:

Enhance the scan status and error reporting from EMA

1. Refactoring the scan status route to make it simpler and separate the overall status from the per-route status.
2. Fix the out of order scan status and scan data messages by changing the event portal endpoint to become a direct component
3. Fix the complete scan status for EMA in offline mode (complete status is persisted in DB)
4. Fix the complete scan status for EMA in offline mode (complete status is logged into the console)
5. Implement the complete status for empty scan types and its descendent in the RouteBundle hierarchy after parsing the hierarchy (no depth constraint) and extract all possible root-to-leaf paths
6. Introduce a new exception handling route at global level.
7. Introduce custom exception classes for Scan status and scan data publisher.
8. Introduce exception handlers for scan status and scan data publisher
9. Improve the unit tests for Scan manager and scan service

Manual + unit tests

N/A

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/EmptyScanEntityProcessorImpl.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/EmptyScanEntityProcessorImpl.java
@@ -1,13 +1,11 @@
 package com.solace.maas.ep.event.management.agent.processor;
 
 import com.solace.maas.ep.event.management.agent.plugin.constants.RouteConstants;
-import com.solace.maas.ep.event.management.agent.plugin.constants.ScanStatus;
 import com.solace.maas.ep.event.management.agent.plugin.processor.EmptyScanEntityProcessor;
 import com.solace.maas.ep.event.management.agent.repository.model.scan.ScanRecipientHierarchyEntity;
 import com.solace.maas.ep.event.management.agent.repository.scan.ScanRecipientHierarchyRepository;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.camel.Exchange;
-import org.apache.camel.ProducerTemplate;
 import org.springframework.stereotype.Component;
 
 import javax.transaction.Transactional;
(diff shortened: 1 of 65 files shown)


service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/EmptyScanEntityProcessorImpl.java
```diff
@@ -0,0 +1,83 @@
...
+                    log.info("Scan request [{}]: Encountered an empty scan type [{}].", scanId, childScanType);
+                    sendCompleteStatusForScanType(scanId, messagingServiceId, childScanType);
+                    log.info("Scan request [{}]: The status of [{}]" + " is: [{}].", scanId, childScanType, ScanStatus.COMPLETE);
+                });
+            }
+        });
+    }
+
+    public void sendCompleteStatusForScanType(String scanId, String messagingServiceId, String scanType) {
+        producerTemplate.send("direct:processScanStatusAsComplete?block=false&failIfNoConsumers=false", exchange -> {
```
slunelsolace, line 77: We shouldn't be using Producer Template inside of Processors.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/ScanStatusPerRouteProcessor.java
```diff
@@ -0,0 +1,68 @@
...
+        topicDetails.put("messagingServiceId", messagingServiceId);
+        topicDetails.put("scanId", scanId);
+        topicDetails.put("status", status.name());
+        topicDetails.put("scanDataType", scanType);
+
+        ScanDataStatusMessage scanDataStatusMessage = new
+                ScanDataStatusMessage(orgId, scanId, status.name(), description, scanType);
+
+        try {
+            scanStatusPublisher.sendScanDataStatus(scanDataStatusMessage, topicDetails);
```
slunelsolace, line 62: Why call this inside of the processor? Can't we use the route instead?

slunelsolace (changes requested):
Can we remove calling ProducerTemplate from Processors? We should be using the Route to forward the body after the Processor sets it.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/EmptyScanEntityProcessorImpl.java
AHabes: Thanks for the feedback. The processor and associated routes have been updated.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/ScanStatusPerRouteProcessor.java
AHabes: Thanks for the feedback. Actually, there are several "publishing processors" that are used to publish the scan data, scan status, scan logs, and import data. If we would like to replace all these processors, then I suggested to implement this change in a separate PR, as it is going to be a bigger task. 

Kindly let me know if this change is necessary in the near future, so I can keep track of it.

slunelsolace (approved):

PR #40, opened 2023-02-07
DATAGO-50106: revert camel library

gregmeldrum:
The new camel library had backward incompatible changes and was causing errors uploading the zip file for import

Reverted the camel library back to 3.16.0

Manually tested. The io exception disappeared

No

### DIFF
--- a/service/application/pom.xml
+++ b/service/application/pom.xml
@@ -15,7 +15,6 @@
     <description>event-management-agent application</description>
 
     <properties>
-        <camel.version>3.19.0</camel.version>
         <springdoc.version>1.6.11</springdoc.version>
         <snakeyaml.version>1.31</snakeyaml.version>
     </properties>
--- a/service/plugin/pom.xml
+++ b/service/plugin/pom.xml
@@ -20,7 +20,6 @@
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <snakeyaml.version>1.31</snakeyaml.version>
-        <camel.version>3.19.0</camel.version>
     </properties>
 
     <dependencies>
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,5 @@ data_collection
 
 application-mysql-*.yml
 .env
+
+*dependency-reduced-pom.xml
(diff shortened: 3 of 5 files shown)


service/application/pom.xml
```diff
@@ -15,7 +15,6 @@
     <description>event-management-agent application</description>
 
     <properties>
-        <camel.version>3.19.0</camel.version>
```
moodiRealist, line 18: Can you please explain the relationships between these pom files and which of them will provide the value for ${camel.version} ?

Can all these pom files get their camel.version from one pom file only? (this question can be invalid depending on the answer to the former question)

gregmeldrum: There is a top-level pom.xml file, and then each directory under the top level has it's own pom.xml. If the subdirectory pom.xml has a parent section, then it inherits the properties of the top-level pom.xml.

In this case the local-storage-plugin/pom does not have a parent section pointing to the top-level pom.xml, so I need to change the version in its pom.xml. The other two subdirectories (application and plugin) inherit from the top-level pom.

@slunelsolace Can we add a parent section to the local-storage-plugin/pom.xml to point to the top-level pom or was that omitted by design?

slunelsolace: @gregmeldrum It's by design. If a project doesn't need the parent dependency it shouldn't be required to import it. Later on these modules should be in their own repositories. I don't think we should be tightly coupling them or decoupling them could be a pain later.

I'd be more in favor are starting an initiative to get plugins out of this repository.

service/plugin/pom.xml
```diff
@@ -20,7 +20,6 @@
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <snakeyaml.version>1.31</snakeyaml.version>
-        <camel.version>3.19.0</camel.version>
```
moodiRealist, line 23: Based on @slunelsolace comment above, would you bring back this explicit dependency for plugins?

gregmeldrum: The plugin directory has the camel infrastructure that a plugin would use, like logging, base processors, and the ability to send data to EP via the eVMR. It's not a specific plugin implementation so I think it makes sense to keep it as is.

moodiRealist (approved):

PR #46, opened 2023-03-02
added the servlet multipart file size property

AHabes:

Increase the manual import file size limit to 5MB.

By updating the `application.yml`

Manually

N/A

### DIFF
--- a/service/application/src/main/resources/application.yml
+++ b/service/application/src/main/resources/application.yml
@@ -28,6 +28,10 @@ spring:
       enabled: true
       settings:
         web-allow-others: true
+  servlet:
+    multipart:
+      max-request-size: 5MB
+      max-file-size: 5MB
   main:
     allow-bean-definition-overriding: true
 


service/application/src/main/resources/application.yml
```diff
@@ -28,6 +28,10 @@ spring:
       enabled: true
       settings:
         web-allow-others: true
+  servlet:
+    multipart:
+      max-request-size: 5MB
+      max-file-size: 5MB
```
moodiRealist, line 34: Would you create a sibling PR on ep-core side to make sure this is generated when creating configs from EP side?

slunelsolace, line 34: Shouldn't we allow this to be configured? 5MB might be way too small for some scans.

AHabes: yeah, that is a separate task and PR. It is already in progress.

moodiRealist (approved):

PR #50, opened 2023-04-04
Fix mixed scanIds in the logs of consecutive scans

AHabes:

To fix the issue of mixed scanIds in the logs of consecutive scans

Issue:

mixed scanIds

By clearing the MDC context upon receiving new scan commands.

Manual

**Fix:**
fixed issue

N/A


moodiRealist (approved):
LGTM.

But let's add before after screenshots of how the results look like (in relation to the bug) to the description field, especially when the PR is about a bug ticket. 
We believe you if you say it works, but for historical purposes (looking at this PR few months in future by a new employee) screenshots are appreciated.

AHabes:
> LGTM.
> 
> But let's add before after screenshots of how the results look like (in relation to the bug) to the description field, especially when the PR is about a bug ticket. We believe you if you say it works, but for historical purposes (looking at this PR few months in future by a new employee) screenshots are appreciated.

@moodiRealist Thanks for the follow up. I've updated the ticket accordingly.

PR #57, opened 2023-05-02
DATAGO-54615:remove all gcr and ecr references from the pom file

amanek-solace:
- simple refactoring of pom file to be consistent with other projects
- renaming the version to 1.0.0-SNAPSHOT
- updated to the latest patch version of spring-boot 2.7 that is supported
- removing aws/gcr references in the pom file because the first phase is to create a jar file only.

- simple pom file changes

- mvn clean compile worked with no issues

- not that i am aware of


gregmeldrum (approved):

shooshmand-sol (approved):
👍

PR #58, opened 2023-05-02
Adding Build Workflow 

shooshmand-sol:
Adding a build workflow, currently it runs PMD/Checkstile/Findbugs/Unit Tests
Buidling the Jar file, SonarQube, Whitesource will be added later

Adding a build workflow to get triggered when we merge into main branch and for PRs

Sample Run: 
https://github.com/SolaceLabs/event-management-agent/actions/runs/4865219741/jobs/8675255781

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,18 @@
+name: Test & Build
+on:
+  push:
+    branches:
+      build_workflow
+jobs:
+  Test-Build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Static Code Analysis
+        run: mvn --batch-mode compile -Pmaas-static-code-analysis --file service/pom.xml
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,7 +11,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up JDK 11
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v3
         with:
           java-version: 11
           cache: 'maven'
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,7 +4,7 @@ on:
     branches:
       build_workflow
 jobs:
-  Test_Build:
+  Run Tests and Build Artifact:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Set up JDK 11
+        uses: actions/setup-java@v3
+        with:
+          java-version: 11
+          distribution: 'temurin'
+          cache: 'maven'
+      - name: Static Code Analysis
+        run: mvn -B compile -Pmaas-static-code-analysis --file service/pom.xml
+      - name: Unit Tests
+        run: mvn -B test -Dcheckstyle.skip -Dpmd.skip -Dcpd.skip -Dfindbugs.skip -Dspotbugs.skip --file service/pom.xml
```
shooshmand-sol, line 25: we do not want to re-run PMD/findbugs test again in this step, since they have been executed in previous step.

amanek-solace: we don't need to to save some cycles in the build but if i remember correctly the sca tools doesn't take a long time to execute.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,25 @@
+name: Build
+on:
+  push:
+    branches:
```
johnvincentcorpuz, line 4: This is a good first step. But would recommend adding exceptions for builds/tests to run on specific paths: i.e https://github.com/SolaceDev/maas-billing-v2/blob/8961d14770a4113486267a95cea288e51009ff3e/.github/workflows/release.yml#L5

johnvincentcorpuz: Or only run build/test when there's a change in the service folder. Example: https://github.com/SolaceDev/maas-billing-v2/blob/8961d14770a4113486267a95cea288e51009ff3e/.github/workflows/build-helm-chart.yml#L10

.github/workflows/build.yaml
shooshmand-sol: But I'm ok to remove the skip stages.

.github/workflows/build.yaml
shooshmand-sol: That's a good idea, so we dont need to run tests on for exampl egithub actions changes.

shooshmand-sol: after discussing it with @amanek-solace we decided to add this later.

johnvincentcorpuz: lol that's a two line change: 
```
  paths:
      - 'service/**'
```

PR #60, opened 2023-05-03
Adding SonarQube scan stage to Build workflow

shooshmand-sol:
Added a step for running SonarQube

Added step to run SonarQube passing required parameters to run
- Since we already executed tests in previous steps, no longer need to run them in this step.

Test Run:
https://github.com/SolaceLabs/event-management-agent/actions/runs/4875826045

I removed the separate workflow which was running SonarQube


johnvincentcorpuz (approved):
👍

PR #61, opened 2023-05-08
DATAGO-54673: Deploy Artifacts To Github Packages

shooshmand-sol:
Deploying Jar artifacts to github packages if all tests/SonarQube pass

- Changes build workflow to deploy to github packages.
- The current flow is
- 1- Static code Analysis
- 2- Unit Tests
- 3- Generate Artifacts
- 4- Sonar
- 5- Deploy

We want SonarQube to pass before `Deploying` them to Github Packages so, we first create Jars in Step 3, and if SonarQube passed, we deploy them; Step three artifacts are also used by SonarQube steps.

Test Run: 
https://github.com/SolaceLabs/event-management-agent/actions/runs/4918913537/jobs/8786059872

- I had to change groupdID `com.solace.maas.plugin.confluentSchemaRegistry` to `confluent-schema-registry` since Github Package expects group ids to be lowercase (see https://github.com/orgs/community/discussions/23474)

- Also renamed `service/confluentSchemaRegistry-plugin` to `service/confluent-schema-registry-plugin` for consistency.
- Removed `nexus.release` references and added Github distribution management section to different POMs


slunelsolace (approved):

johnvincentcorpuz (approved):
Pattern looks good to me to get it off the ground by being able to run static code/sonar, unit tests and eventually deploy the package.

There's a couple of follow questions
- When do we publish artifacts to the package registry is it only during merges to the `main` branch or even for PR branches?
- When we deploy a package on merges to the `main` branch. Should we consider removing `-SNAPSHOT`? Or it's only during the release that we drop the `-SNAPSHOT` name?
- We should start looking into a versioning strategy. For example, each and every merge to the `main` branch are we required to bump the version or we keep the same versioning number? 
- We should also be able to trace the package versions easily to a git commot.

johnvincentcorpuz:
We also need to set the `Squash commits` here on merges to clean up the commit history a little bit

service/application/pom.xml
```diff
@@ -157,8 +157,8 @@
             <version>1.0.0-SNAPSHOT</version>
         </dependency>
         <dependency>
-            <groupId>com.solace.maas.plugin.confluentSchemaRegistry</groupId>
-            <artifactId>confluentSchemaRegistry-plugin</artifactId>
+            <groupId>com.solace.maas.plugin.confluent-schema-registry</groupId>
```
johnvincentcorpuz, line 160: 👍 consistency yes!

PR #63, opened 2023-05-09
Add concurrentcy/timeout and path filtering to build workflow

shooshmand-sol:
Improving build workflow to:
- have a timeout of 20 minutes
- disable concurrent runs for the same PR (or main branch)
- not triggering the build for changes in .github folder

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,9 +3,9 @@ on:
   push:
     branches:
       - main
-      paths-ignore:
-        - '.gitignore'
-        - '.github/**'
+    paths-ignore:
+      - '.gitignore'
+      - '.github/**'
   pull_request:
     types: [opened, synchronize, reopened]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -2,14 +2,22 @@ name: Build
 on:
   push:
     branches:
-      main
+      - main
+    paths-ignore:
+      - '.gitignore'
```
johnvincentcorpuz, line 7: Noting that we are running unit-tests here for other Doc/README changes. I.e it will generate a JAR file on pushes to the main branch. If that's the agreed upon process that we run a full build on all changes, then Im okay with this.

johnvincentcorpuz (approved):
Approving the changes, with comments about the running a full build on changes that should not trigger a new jar artifact.

.github/workflows/build.yaml
shooshmand-sol: That's a good point, I will add ".md" files to the ignore list.

PR #65, opened 2023-05-12
Add test reporters

shooshmand-sol:
Adding test reporters of static code analysis and unit tests to build workflow

Following actions have been used to publish test results
- jwgmeligmeyling/checkstyle-github-action@master
- jwgmeligmeyling/pmd-github-action@master
- jwgmeligmeyling/spotbugs-github-action@master
- dorny/test-reporter@v1
- 
https://github.com/SolaceLabs/event-management-agent/actions/runs/4982607106

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -58,70 +58,70 @@ jobs:
         run: mvn -B compile -Pmaas-static-code-analysis --file service/pom.xml
       - name: Unit Tests
         run: mvn -B test $SKIP_FLAGS_NON_UNIT_TESTS --file service/pom.xml
-#      - name: Generate Artifacts
-#        run: |
-#          mvn install $SKIP_FLAGS_ALL_TESTS --file service/pom.xml
-#      - name: Sonar Scan
-#        env:
-#          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-#          SONAR_HOST_URL: ${{ secrets.SONAR_HOST_URL }}
-#        run: |
-#          mvn -B $SKIP_FLAGS_ALL_TESTS \
-#            org.sonarsource.scanner.maven:sonar-maven-plugin:sonar -Dsonar.projectKey=SolaceLabs_runtime-agent \
-#            --file service/pom.xml
-#      - name: WhiteSource Scan
-#        if: env.WHITESOURCE_SCAN=='true'
-#        env:
-#          unified_agent_url: "https://github.com/whitesource/unified-agent-distribution/releases/latest/download/wss-unified-agent.jar"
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -20,12 +20,17 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20
     permissions:
-      contents: read
-      packages: write
+       contents: read
+       packages: write
+       pull-requests: write
+       actions: write
```
shooshmand-sol, line 26: required for plugins t o works, so they are able to set status checks and write comments to PRs

.github/workflows/build.yaml
```diff
@@ -101,3 +106,33 @@ jobs:
...
+        if: always()
+        uses: jwgmeligmeyling/pmd-github-action@master
+        with:
+          path: '**/pmd.xml'
+      - name: Publishing Test Results - SpotBugs
+        if: always()
+        uses: jwgmeligmeyling/spotbugs-github-action@master
+        with:
+          path: '**/spotbugsXml.xml'
+      - name: Publishing Test Results - Unit Tests Pre-Condition
```
shooshmand-sol, line 124: added to check that unit tests reports exist, otherwise unit test report fail

.github/workflows/build.yaml
```diff
@@ -101,3 +106,33 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           mvn deploy $SKIP_FLAGS_ALL_TESTS -Dmaven.install.skip=true --file service/pom.xml
+      - name: Publishing Test Results - Checkstyle
+        if: always()
```
shooshmand-sol, line 110: `always()` means to run this step `always` even if previous steps fail (which is the xecpected behaviour, we want to see the test report even if tests fail)

.github/workflows/build.yaml
```diff
@@ -20,12 +20,17 @@ jobs:
...
+       packages: write
+       pull-requests: write
+       actions: write
+       statuses: write
+       checks: write
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
```
johnvincentcorpuz, line 33: Will this work on the `main` branch?

shooshmand-sol: should work based on the similar pattern we had for SonarQube checkouts: https://github.com/SolaceDev/maas-ep-core/blob/1f97c227869366fce9476f7e9bf1488c69d70e4f/.github/workflows/sonar-scan.yml#L15

johnvincentcorpuz (approved):
Looks good to me. Thanks!

PR #68, opened 2023-05-19
Add the traceId to all messages streamed by EMA to EP

AHabes:

To add the traceId to all EMA's messages.

By integrating the traceId into the scan request and all message types.

**Update**: The manual import routes and processors have been updated to integrate the traceId for manual import jobs. Also, new log messages have been added to reflect the status of manual import jobs.

manual import logs

Manual

N/A


service/application/src/main/java/com/solace/maas/ep/event/management/agent/service/ScanService.java
```diff
@@ -244,16 +244,6 @@ public Optional<ScanEntity> findById(String scanId) {
         return repository.findById(scanId);
     }
 
-    protected Exchange scan(String groupId, String scanId, RouteEntity route,
```
AHabes, line 247: This old scan method has been removed since it is not used by any other methods/classes

slunelsolace (approved):

PR #70, opened 2023-05-23
Add release workflow

shooshmand-sol:
Adding release workflow.
The workflow has two inputs:
- releaseVersion: version that we want to deploy in x.y.z format
- developmentVersion: next development version (in x.y.z-snapshot format)

Added a workflow file: `.github/workflows/release.yaml` which can be triggered manually.
- Currently it only contians a pre-check to make sure that the inoput version has not been previously released.
The workflow uses github api to query previous released.

Test Runs (in a test repo) for a previously released verion
image

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,7 +18,6 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-          token: ${{ secrets.BUILD_SVC_PAT }}
       - name: Set up JDK 11
         uses: actions/setup-java@v3
         with:
@@ -30,7 +29,7 @@ jobs:
            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           gh api --method GET /repos/${{github.repository}}/releases -f sort=updated -f direction=asc > releases.json
-          release_version_exists=$(jq -r --arg RELEASE_VERIOSN v${{ github.event.inputs.releaseVersion }} '.[].name|select(.|test($RELEASE_VERIOSN))'  releases.json)
+          release_version_exists=$(jq -r --arg RELEASE_VERSION v${{ github.event.inputs.releaseVersion }} '.[].name|select(.|test($RELEASE_VERSION))'  releases.json)
           if [[ ! -z "$release_version_exists" ]]; then
                 echo "Version ${{ github.event.inputs.releaseVersion }} has been previously released. Please change release version."
                 exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,39 @@
...
+        uses: actions/setup-java@v3
+        with:
+          java-version: 11
+          distribution: 'temurin'
+          cache: 'maven'
+      - name: Pre-Release Check - Version
+        env:
+           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh api --method GET /repos/${{github.repository}}/releases -f sort=updated -f direction=asc > releases.json
```
shooshmand-sol, line 32: jq and github cli (gh) are available on github ubuntu runners.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,39 @@
...
+          release_version_exists=$(jq -r --arg RELEASE_VERIOSN v${{ github.event.inputs.releaseVersion }} '.[].name|select(.|test($RELEASE_VERIOSN))'  releases.json)
```
shooshmand-sol, line 33: This command checks whether the release version exists in the ist of releases or not.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,39 @@
...
+        default: "X.Y.Z-SNAPSHOT"
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.BUILD_SVC_PAT }}
```
johnvincentcorpuz, line 21: What's this new token?

.github/workflows/release.yaml
johnvincentcorpuz, line 33: `VERIOSN` typo.

.github/workflows/release.yaml
shooshmand-sol: This is for being able to push changes in pom files without approval, actually not needed for this PR.

shooshmand-sol: I remove it for now :)

.github/workflows/release.yaml
shooshmand-sol: Thanks, updated!

antonysavio-sol (approved):

johnvincentcorpuz (approved):
Changes look good to me. Nice touch on adding a check if the release already exists.

PR #79, opened 2023-05-26
DATAGO-55462: Make sure no critical/high whitesource vulnurability exists

shooshmand-sol:
Checkiing no critical/high whitesource vulnurability exists before release

- Python code added .github/workflows/release_scripts/whitesource_vulnurability_checker.py which queries whitesource for any high/critical vulnurabilities, and if any found fails the workflow.
- Calling Python code added to the release workflow

https://github.com/SolaceLabs/event-management-agent/actions/runs/5092482216/jobs/9153881210
image


.github/workflows/release.yaml
```diff
@@ -36,3 +36,19 @@ jobs:
           else
                 echo "New version: ${{ github.event.inputs.releaseVersion }} going to be released!"
           fi
+      - name: Set up Python 3.8
```
johnvincentcorpuz, line 39: you can add some caching here too.

.github/workflows/release.yaml
```diff
@@ -36,3 +36,19 @@ jobs:
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+      - name: Pre-Release Check - Whitesource vulnurabilities
+        env:
+          WS_APIKEY: ${{ secrets.WHITESOURCE_API_KEY }}
+          WS_PROJECTTOKEN: ${{ secrets.WHITESOURCE_PROJECT_TOKEN }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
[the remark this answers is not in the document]
johnvincentcorpuz: can you we store this new secret in Vault or LastPass for future use that needs dynamoDB access only. I imagine we can use this for most of our vulnerability scanning tools

.github/workflows/release_scripts/whitesource_vulnurability_checker.py
```diff
@@ -0,0 +1,59 @@
...
+        'projectToken': WS_PROJECT_TOKEN
+    }
+    headers = {'content-type': 'application/json'}
+    ws_vulnerability_report_response = requests.request(
+        'post',
+        WS_REST_API_URL,
+        data=json.dumps(ws_payload),
+        headers=headers)
+
+    project_ws_vulnerabilities = json.loads(ws_vulnerability_report_response.text)
```
antonysavio-sol, line 38: As a suggestion , would ws_vulnerability_report_response.json() work here if the content-type is application/json ?

johnvincentcorpuz (approved):
Code looks good. made recommendations for caching (non blocking) and also adding references to our secrets in Last Pass or Vault for future use.

.github/workflows/release_scripts/whitesource_vulnurability_checker.py
```diff
@@ -0,0 +1,59 @@
...
+import json
+import os
+import boto3
+
+WS_REST_API_URL = 'https://saas.whitesourcesoftware.com/api'
+WS_API_KEY = os.getenv('WS_APIKEY')
+WS_PROJECT_TOKEN = os.getenv('WS_PROJECTTOKEN')
+WS_BLOCKING_VULNERABILITIES = "critical,high"
+
+def determine_any_exclusions():
```
antonysavio-sol, line 11: suggestion to rename this method to something like get_exclusions() or get_excluded_libraries() since it just gets exclusion entries and returns a list of ,ie, there is no rule or criteria to determine exclusions here

antonysavio-sol (approved):
Approving , added some comments for small improvements

.github/workflows/release.yaml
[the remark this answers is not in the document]
shooshmand-sol: added to EMA_AWS_KEYS in LastPass in Shared-sc-build-leads folder

.github/workflows/release_scripts/whitesource_vulnurability_checker.py
shooshmand-sol: sounds good, replaced it! thanks

PR #82, opened 2023-05-31
DATAGO-55072: Create Release pipeline for Event Management Agent Jar file 

shooshmand-sol:
Adding maven release steps to release pipeline

- in .github/workflows/release.yaml workflow, added steps for changelog, mvn release and create a release
- in service/pom.xml added maven release plugin and other configs for the release/

Test release created: https://github.com/SolaceLabs/event-management-agent/releases/tag/v1.0.0

will delete this release later.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,6 +18,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          token: ${{ secrets.RELEASE_TOKEN }}
       - name: Set up JDK 11
         uses: actions/setup-java@v3
         with:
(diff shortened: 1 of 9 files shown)


service/pom.xml
```diff
@@ -310,6 +312,15 @@
                         </execution>
                     </executions>
                 </plugin>
+                <plugin>
+                    <artifactId>maven-release-plugin</artifactId>
+                    <version>3.0.0-M1</version>
+                    <configuration>
+                        <tagNameFormat>v@{project.version}</tagNameFormat>
+                        <arguments>-Dmaven.test.skip=true -Dcheckstyle.skip -Dpmd.skip -Dcpd.skip -Dfindbugs.skip -Dspotbugs.skip</arguments>
```
shooshmand-sol, line 320: maven release configs, not running any tests.

service/pom.xml
```diff
@@ -310,6 +312,15 @@
...
+                        <scmCommentPrefix>[ci skip]</scmCommentPrefix>
```
shooshmand-sol, line 321: when maven updates pom files as part of the release we do not want to run the `Build` workflow; This config adds a `[ci skip]` message in the commit that maven release pushes, and since it has `[ci skip]` github will not tirgger any workflow.
image

.github/workflows/release.yaml
```diff
@@ -53,3 +54,27 @@ jobs:
...
+      - name: Set Release Configs
+        run: |
+          export SKIP_FLAGS_NON_UNIT_TESTS="-Dcheckstyle.skip -Dpmd.skip -Dcpd.skip -Dfindbugs.skip -Dspotbugs.skip"
+          echo "SKIP_FLAGS_NON_UNIT_TESTS=$SKIP_FLAGS_NON_UNIT_TESTS" >> $GITHUB_ENV
+          echo "SKIP_FLAGS_ALL_TESTS=$SKIP_FLAGS_NON_UNIT_TESTS -Dmaven.test.skip=true" >> $GITHUB_ENV
+          git config user.email "actions@github.com"
+          git config user.name "GitHub Actions"
+      - name: Maven Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
johnvincentcorpuz, line 66: Just wanted to double check that this token has limited permissions. Thanks,

.github/workflows/release.yaml
```diff
@@ -53,3 +54,27 @@ jobs:
...
+        run: mvn release:prepare release:perform -B --file service/pom.xml -DreleaseVersion=${{ github.event.inputs.releaseVersion }} -DdevelopmentVersion=${{ github.event.inputs.developmentVersion }}
+      - name: Changelog
```
johnvincentcorpuz, line 68: Can you confirm what's the difference between the changelog here on this action, and the next step

johnvincentcorpuz: This one generates the change log, and you feed it in to the next action 👍

shooshmand-sol: The first one generates the change log, but does not add it to the release; we genrate the change log body in this step and pass it to the next action as the body that needs to be added to the generated release.

.github/workflows/release.yaml
shooshmand-sol: This is the token provided by github it self and only has access to this repo.

PR #85, opened 2023-06-05
DATAGO-55072: Change SonarQube Project Key

shooshmand-sol:
Changing SonarQube project key to `SolaceLabs_event-management-agent`
to match what we have in SonarQube

`sonar.projectKey` parameter updated in build workflow


johnvincentcorpuz (approved):
👍

PR #86, opened 2023-06-07
Using SSH keys for release

shooshmand-sol:
Using ssh key for release instead of token

- ssh key pair generated and added as deployment key/secret
- script added to configure the ssh key and update the maven settings file to use the generated key
- pom file added to use ssh for communicating with github

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -55,12 +55,13 @@ jobs:
           export VIRTUAL_ENV=./venv
           python3.8 -m venv $VIRTUAL_ENV && source $VIRTUAL_ENV/bin/activate
           cd ./.github/workflows/release_scripts/  && pip install --quiet -r requirements.txt && python3.8 whitesource_vulnurability_checker.py
-      - name: Prepate Maven Settings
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -55,16 +55,19 @@ jobs:
           export VIRTUAL_ENV=./venv
           python3.8 -m venv $VIRTUAL_ENV && source $VIRTUAL_ENV/bin/activate
           cd ./.github/workflows/release_scripts/  && pip install --quiet -r requirements.txt && python3.8 whitesource_vulnurability_checker.py
+      - name: Prepate Maven Settings
```
rprasad-solace, line 58: I assume this was supposed to say:
```suggestion
      - name: Prepare Maven Settings
```

johnvincentcorpuz (approved):
Looks good to me, aside from Rohan's suggestion to correct the typo.

.github/workflows/release.yaml
shooshmand-sol: Thanks Rohan, updated.

PR #90, opened 2023-06-14
Checking Prisma Vulnurabilites before release

shooshmand-sol:
Making sure that no high/critical Prisma vulnurabilites exist in docker image before release

.github/workflows/release_scripts/prisma_vulnurability_checker.py added to query the latest scan result for EMA docker image. If any critical/high vulnurability found, the script fails.

https://github.com/SolaceLabs/event-management-agent/actions/runs/5270739012/jobs/9530640345

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -20,24 +20,24 @@ jobs:
         with:
           fetch-depth: 0
 
-#      - name: Set up JDK 11
-#        uses: actions/setup-java@v3
-#        with:
-#          java-version: 11
-#          distribution: 'temurin'
-#          cache: 'maven'
-#      - name: Pre-Release Check - Version
-#        env:
-#           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-#        run: |
-#          gh api --method GET /repos/${{github.repository}}/releases -f sort=updated -f direction=asc > releases.json
-#          release_version_exists=$(jq -r --arg RELEASE_VERSION v${{ github.event.inputs.releaseVersion }} '.[].name|select(.|test($RELEASE_VERSION))'  releases.json)
-#          if [[ ! -z "$release_version_exists" ]]; then
-#                echo "Version ${{ github.event.inputs.releaseVersion }} has been previously released. Please change release version."
-#                exit 1
-#          else
-#                echo "New version: ${{ github.event.inputs.releaseVersion }} going to be released!"
-#          fi
+      - name: Set up JDK 11
+        uses: actions/setup-java@v3
+        with:
+          java-version: 11
+          distribution: 'temurin'
+          cache: 'maven'
+      - name: Pre-Release Check - Version
+        env:
+           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh api --method GET /repos/${{github.repository}}/releases -f sort=updated -f direction=asc > releases.json
+          release_version_exists=$(jq -r --arg RELEASE_VERSION v${{ github.event.inputs.releaseVersion }} '.[].name|select(.|test($RELEASE_VERSION))'  releases.json)
+          if [[ ! -z "$release_version_exists" ]]; then
+                echo "Version ${{ github.event.inputs.releaseVersion }} has been previously released. Please change release version."
+                exit 1
+          else
+                echo "New version: ${{ github.event.inputs.releaseVersion }} going to be released!"
+          fi
       - name: Set up Python 3.8
         uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -47,14 +47,27 @@ jobs:
         env:
           WS_APIKEY: ${{ secrets.WHITESOURCE_API_KEY }}
           WS_PROJECTTOKEN: ${{ secrets.WHITESOURCE_PROJECT_TOKEN }}
-          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.EMA_AWS_ACCESS_KEY_ID }}
```
shooshmand-sol, line 50: changed whitesource to use `EMA_AWS_ACCESS_KEY_ID` secret, which is only available in `prod` environment.

.github/workflows/release_scripts/prisma_vulnurability_checker.py
```diff
@@ -0,0 +1,90 @@
...
+    whitesource_exclusion_entries = whitesource_exclusion_table.scan()['Items']
+
+    exclusions = set()
+    for exclusion in whitesource_exclusion_entries:
+        exclusions.add(exclusion['package'])
+
+    return exclusions
+
+
+def block_print(long_string, each_line_length=75):
```
shooshmand-sol, line 31: utility function, instead of prints long lines, this function split lines and prints them so they are easier to read.
specially useful for vulnurabilty descriptions.
example output:

```
In FasterXML jackson-databind before 2.14.0-rc1, resource
	exhaustion can occur because of a lack of a check in primitive value
	deserializers to avoid deep wrapper array nesting, when the
	UNWRAP_SINGLE_VALUE_ARRAYS feature is enabled. Additional fix version in
	2.13.4.1 and 2.12.17.1
```

instead of a single line output!

.github/workflows/release_scripts/prisma_vulnurability_checker.py
```diff
@@ -0,0 +1,90 @@
...
+
+def find_all_high_critical_vulnerabilities_to_resolve(excluded_libraries):
+    # Authenticate
+    authenticate_payload = {
+        'username': PRISMA_ACCESS_KEY,
+        'password': PRISMA_ACCESS_KEY_SECRET
+    }
+    authenticate_headers = {'Accept': HTTP_JSON_HEADER, 'Content-type': HTTP_JSON_HEADER}
+
+    get_token_request = requests.post(
```
johnvincentcorpuz, line 43: Since it will only make one call to prisma cloud to get the results, im good with generating a session token per call in this scenario

.github/workflows/release.yaml
```diff
@@ -47,14 +47,27 @@ jobs:
...
+          AWS_DEFAULT_REGION: ${{ secrets.EMA_AWS_DEFAULT_REGION }}
         run: |
           pip install --quiet --upgrade pip
           export VIRTUAL_ENV=./venv
           python3.8 -m venv $VIRTUAL_ENV && source $VIRTUAL_ENV/bin/activate
           cd ./.github/workflows/release_scripts/  && pip install --quiet -r requirements.txt && python3.8 whitesource_vulnurability_checker.py
+      - name: Pre-Release Check - Prisma vulnurabilities
+        env:
+          PRISMA_ROOT_API_URL: ${{ secrets.PRISMA_ROOT_API_URL }}
+          DOCKER_IMAGE_TO_CHECK: ${{ secrets.PRISMA_DOCKER_IMAGE_TO_CHECK }}
```
johnvincentcorpuz, line 61: How is the docker image tag stored in a secret? i.e is this image tag always changing?

shooshmand-sol: It's always checking the latest image with `main` tag. Just didn't want to expose our ecr address in the code!
We only need to scan `main` tg, since during release we always release the latest main image.

.github/workflows/release_scripts/prisma_vulnurability_checker.py
shooshmand-sol: exactly, for a signle call it's ok, if we want to query several repos, we should generate the token once and re-use the previously generate token in following calls.

.github/workflows/release.yaml
johnvincentcorpuz: that's really good were not exposing it. Good thinking.

johnvincentcorpuz (approved):
👍

PR #91, opened 2023-06-15
Push EMA main image with release tag to ECR

shooshmand-sol:
At the end of release, push the main docker image with tag `release_version` (input of the release workflow) to ECR

Added a step to pull/tag the main image and push it to ECR.

https://github.com/SolaceLabs/event-management-agent/actions/runs/5282947289/jobs/9558528560

I also commented PRisma check step until we raise Prisma Jiras

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -109,7 +109,7 @@ jobs:
         run: |
           MANIFEST=$(aws ecr batch-get-image --repository-name ${{ github.event.repository.name }} --image-ids imageTag=main --query 'images[].imageManifest' --region ${{ secrets.EMA_AWS_DEFAULT_REGION }} --output text)
           echo $MANIFEST
-#          aws ecr put-image --repository-name ${{ github.event.repository.name }} --image-tag ${{ github.event.inputs.releaseVersion }} --image-manifest $MANIFEST --region ${{ secrets.EMA_AWS_DEFAULT_REGION }}
+          aws ecr put-image --repository-name ${{ github.event.repository.name }} --image-tag ${{ github.event.inputs.releaseVersion }} --image-manifest \"\$MANIFEST\" --region ${{ secrets.EMA_AWS_DEFAULT_REGION }}
 #          
 #          ECR_REPO="${{ steps.login-ecr.outputs.registry }}/"
 #          MAIN_IMAGE="$ECR_REPO:main"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -96,3 +96,21 @@ jobs:
...
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1.6.0
+      - name: ECR Docker Image Release
+        run: |
+          ECR_REPO="${{ steps.login-ecr.outputs.registry }}/${{ github.event.repository.name }}"
+          MAIN_IMAGE="$ECR_REPO:main"
+          RELEASE_IMAGE="$ECR_REPO:${{ github.event.inputs.releaseVersion }}"
+          echo "Release image : $RELEASE_IMAGE" 
+          docker pull $MAIN_IMAGE
```
johnvincentcorpuz, line 114: It's a one off task during release, but typically if it's coming from ECR it's better to use ECR's update tag thru the CLI than pulling the image and retagging it. Pulls depending on how big the image is, can be expensive

johnvincentcorpuz: This is a snippet we use for Jenkins when we tag an image, we don't pull it, but rather get the manifest metadata and then add an image tag based on the metadata.
```
MANIFEST = $("aws ecr batch-get-image --repository-name ${repository} --image-ids imageTag=${image_Tag} --query 'images[].imageManifest' --region us-east-1 --output text")
 aws ecr put-image --repository-name ${repository} --image-tag ${tag} --image-manifest \"\$MANIFEST\" --region us-east-1
```

shooshmand-sol: That's a great idea, going to use this approach.

.github/workflows/release.yaml
```diff
@@ -55,19 +55,19 @@ jobs:
...
-          PRISMA_ACCESS_KEY: ${{ secrets.PRISMA_ACCESS_KEY }}
-          PRISMA_ACCESS_KEY_SECRET: ${{ secrets.PRISMA_ACCESS_KEY_SECRET }}
-          AWS_ACCESS_KEY_ID: ${{ secrets.EMA_AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.EMA_AWS_SECRET_ACCESS_KEY }}
-          AWS_DEFAULT_REGION: ${{ secrets.EMA_AWS_DEFAULT_REGION }}
-        run: |
-          export VIRTUAL_ENV=./venv
-          python3.8 -m venv $VIRTUAL_ENV && source $VIRTUAL_ENV/bin/activate
-          cd ./.github/workflows/release_scripts/  && python3.8 prisma_vulnurability_checker.py
+#      - name: Pre-Release Check - Prisma vulnurabilities
```
johnvincentcorpuz, line 58: Commenting that this will be disabled, until communicated to the correct squad.

johnvincentcorpuz (approved):
👍

PR #101, opened 2023-07-17
Initial drop Solace SEMPv2 Config Plugin

195858:

 Introduces a generated SEMPv2 REST Client and a Solace Config Plugin.

The Solace config plugin receives SEMPv2 commands - a combination of desired state and SEMPv2 payload describing the broker resource to create/update/delete.

The config plugin implements tasks similar to an sensible task that encapsulates the required GET/POST/PATCH/DELETE operations to achieve the desired ABSENT/PRESENT state.

Generated SEMPv2 REST client (from OpenAPI)
Config plugin aligned with plugin framework and introducing a simple task framework.
Changes to EMA so it can process commands and also resolve Hava type when parsing JSON more dynamically.

Integration/JUnit tests in both config plugin and EMA agent projects.

Changes to core classes especially Jackson JSON deserialisation


service/application/src/main/resources/application.yml
```diff
@@ -68,21 +68,48 @@ eventPortal:
...
+      standalone: false
       rtoSession: false
       enableHeartbeats: false
       testHeartbeats: false
       connections:
         - name: eventPortalGateway
-          authenticationType: ${EP_GATEWAY_AUTH}
-          url: ${EP_GATEWAY_URL}
-          msgVpn: ${EP_GATEWAY_MSGVPN}
+          authenticationType: ${EP_GATEWAY_AUTH:basicAuthentication}
```
gregmeldrum, line 77: Please revert this file to the original content.

gregmeldrum:
I'm having trouble reviewing this PR, my web pages freezes since there are so many files. I'll muddle my way through. 
When I tried to do a mvn build from the command line it failed with static analysis errors. Can you please fix them:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:check (default) on project plugin: Failed during checkstyle execution: There are 5 errors reported by Checkstyle 8.19 with resources/checkstyle/checkstyle.xml ruleset. -> [Help 1]
```

thrillstone:
@195858 I assume we've compiled a client from the SEMP v2 swagger spec in this PR. Wouldn't it make more sense to not check in the compiled client but instead have mvn generate the sources as a part of the build? We don't want to have to change so many files with each new version of SEMP v2... plus things like this PR being so big wouldn't be a problem.

195858:
> @195858 I assume we've compiled a client from the SEMP v2 swagger spec in this PR. Wouldn't it make more sense to not check in the compiled client but instead have mvn generate the sources as a part of the build? 

That makes more sense and I have changed the maven module to generate the code when the module is built.

195858:
@gregmeldrum - have merged your changes and also changed the SEMPv2 client module to generate code on build.

Does this resolve the issues?

service/application/src/main/resources/application.yml
195858: Done

PR #107, opened 2023-08-01
DATAGO-59449: Keep track of release versions in manifest db

shooshmand-sol:
Keep track of release versions in manifest db

Added a script to update the manifest table

Test run:
https://github.com/SolaceLabs/event-management-agent/actions/runs/5730985616


johnvincentcorpuz (approved):
👍

PR #108, opened 2023-08-02
DATAGO-59496: Make docker run for user without root access

gregmeldrum:

Two changes:
1. Change docker base image to eclipse-terumin to be the same as the rest of the microservices (this addresses https://sol-jira.atlassian.net/browse/DATAGO-59217)
2. Make the docker container run for users without root access in docker

1. The base image was updated in the `buildEventManagementAgentDocker.sh` script to manually set the base image to eclipse-terumin.
2. The base image directory was removed since it's no longer used.
3. The Docker uses /opt/ema as it's new working directory since it can be used by docker containers running without root access.

Docker container was manually built and testing in the following scenarios:
1. Run a scan, connected to EP with docker running with root user
2. Run a scan, connected to EP with docker running as non root user
3. Run a scan and download, not connected to EP with docker running with root user
4. Run a scan and download, not connected to EP with docker running as non root user

No


service/application/src/main/resources/logback-spring.xml
```diff
@@ -35,7 +35,7 @@
     </appender>
     <appender name="StreamingAppender" class="com.solace.maas.ep.event.management.agent.logging.StreamingAppender"/>
 
-    <springProfile name="default,mysql,mysql-dev,DEV,TEST">
+    <springProfile name="default,mysql,mysql-dev,DEV,TEST,greg1,gameday">vi d
```
slunelsolace, line 38: @gregmeldrum Should these profiles be pushed to the main branch?

slunelsolace (approved):

PR #112, opened 2023-08-11
DATAGO-59967: Move the EMA to JDK 17 + Spring 3.1.2

gregmeldrum:
There are high vulnerabilities in the dependencies of Spring 2.7.6 that we cannot fix without upgrading Spring. We decided to go to Spring 3.1.2 and as part of the change, we need to upgrade to JDK 17.

This PR is an attempt to maintain EMA feature parity while updating our dependencies to newer versions.

After updating the Spring libraries from 2.7.6 to 3.1.2 I also needed to update other dependencies. For instance, Camel went from 3.16.0 to 4.0.
Part of the move also involved changing the javax package to jakarta.

A full regression on the EMA:
Kafka broker scan
Confluent schema registry scan
Solace broker scan
online scan from EP
offline scan from REST
offline get zip file from REST
online push scan to EP from REST

No


service/application/src/test/java/com/solace/maas/ep/event/management/agent/plugin/route/handler/base/AsyncRoutePublisherImplTests.java
```diff
@@ -44,7 +45,10 @@ public class AsyncRoutePublisherImplTests {
     public void testStart() {
         AsyncWrapper asyncWrapper = mock(AsyncWrapper.class);
 
-        Exchange exchange = ExchangeBuilder.anExchange(mock(ExtendedCamelContext.class))
+        CamelContext cc = mock(CamelContext.class);
```
gregmeldrum, line 48: ExtendedCamelContext previously extend CamelContext. In the 3 -> 4 camel migration they broker this, so now the ExtendedCamelContext is a member of the CamelContext. Needed to update the tests to reflect this.

service/resources/pmd/pmd-rules.xml
```diff
@@ -13,12 +13,14 @@
...
         <exclude name="GuardLogStatement"/>
         <exclude name="JUnit4TestShouldUseBeforeAnnotation"/>
+        <exclude name="JUnit5TestShouldBePackagePrivate"/>
         <exclude name="JUnitAssertionsShouldIncludeMessage"/>
         <exclude name="JUnitTestContainsTooManyAsserts"/>
         <exclude name="JUnitTestsShouldIncludeAssert"/>
         <exclude name="PreserveStackTrace"/>
         <exclude name="SwitchStmtsShouldHaveDefault"/>
         <exclude name="UseVarargs"/>
+        <exclude name="LiteralsFirstInComparisons"/>
```
gregmeldrum, line 23: This rule makes no sense. It was picking up code like this:

```
assertTrue(thrown.getMessage().contentEquals("Some value"));
```

There is no opportunity to do a literal first comparison in this case.
Excluding.

service/solace-plugin/src/test/java/com/solace/maas/ep/event/management/agent/plugin/solace/processor/SolaceSempClientTests.java
```diff
@@ -135,6 +139,46 @@ public void badConnectionUrlTest() throws IOException {
         assertEquals("Could not construct URI from worst url ever", exception.getMessage());
     }
 
+    @Test
```
gregmeldrum, line 142: Added coverage for the if check above.

gregmeldrum:
Not sure why SonarQube is failing. I have 100% coverage in the file in question when running locally.

slunelsolace:
> Not sure why SonarQube is failing. I have 100% coverage in the file in question when running locally.

I'll take a look on Monday.

slunelsolace (approved):

PR #122, opened 2023-10-02
Use SolaceProducts GitHub Packages Repository

shooshmand-sol:
Changing refrences to GitHub Packages repositories to SolaceProducts

Updating pom files to use `https://maven.pkg.github.com/SolaceProducts/event-management-agent` instead of
`https://maven.pkg.github.com/SolaceLabs/event-management-agent`


rprasad-solace (approved):

antonysavio-sol (approved):
👍

PR #124, opened 2023-10-10
add EMA version to heartbeat message

pwang1997:

Add EMA version to heartbeat message for debugging purposes.

Add EMA version to heartbeat message.

Update test 

This change will be minor upversioned.

### DIFF
--- a/service/application/src/test/java/com/solace/maas/ep/event/management/agent/mop/MopTests.java
+++ b/service/application/src/test/java/com/solace/maas/ep/event/management/agent/mop/MopTests.java
@@ -52,7 +52,7 @@ public class MopTests {
 
     @Test
     public void testHeartbeatAgentMOPProtocol() {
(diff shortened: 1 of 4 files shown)


service/application/src/test/java/com/solace/maas/ep/event/management/agent/mop/MopTests.java
```diff
@@ -52,7 +52,7 @@ public void ack() {
 
     @Test
     public void testHeartbeatAgentMOPProtocol() {
-        HeartbeatMessage message = new HeartbeatMessage("runtimeId", "timestamp");
+        HeartbeatMessage message = new HeartbeatMessage("runtimeId", "timestamp", "1.1.2");
```
moodiRealist, line 55: Since this change is not backwardCompatible, we should upversion ema on the minor version

moodiRealist (approved):

PR #125, opened 2023-10-17
DATAGO-63031: Fix logging of traceId, actorId and orgId so it's consistent between EMA and EP

gregmeldrum:

1. When a scan is initiated from EP, EP logs the messages received from the EMA as well as activities caused by those messages. We want the orgId, traceId and actorId in the EP logs to be the same for the EP initiated logs, as well as the logs caused by messages received by the EMA. To accomplish this, orgId, traceId and actorId must be passed along the entire flow from EP, through the EMA and then back to EP again. This way, all logs (even on the EMA) will contain the same data and be easier to correlate.

2. When a scan import is initiated on the EMA, we similarly want the traceId to be consistent for all of the logs in EP and EMA in order to make it easier to see the entire stream of logs for the scan import action. In the case of import, we don't set the actorId, since the EMA REST API is unauthenticated, so there is no way to determine the user.

Actor id was already added to the MOPMessage in maas-base. As part of this change, it is being added to the EMA's MOPMessage structure as well so it is available to all messages.

In EP, the actor Id is set on the scan command message sent to the EMA. The EMA maintains this actorId through all of the routes, and sends it back to EP in the response messages so EP can add it to the logging context.

In the EMA, the actorId is added to the MDCContext and exchange headers wherever this is done from the traceId.

For scan import operations, we want to set a traceId on the EMA. Since sleuth is no longer supported with the move to spring boot 3.0, micrometer was added as the tracing framework. The micrometer generated traceId gets used in the EMA logs, as well as gets sent to EP, so it can be used as the traceId for the EP logs as well.

The RouteConstant for the trace id is changed to "traceId" since this is what micrometer uses and we get this added to our logs for free.

The following operations were tested:
**Functionality**
* Scan initiated from EP
* Scan initiated from EMA
* File download initiated from EMA
* Scan import initiated from EMA

**Backward Compatability**
* Scan initiated from EP using an old EMA
* Scan initiated from an old EMA to EP

Micrometer generates 128 bit trace ids which are twice as long as what sleuth generates.

### DIFF
--- a/service/application/pom.xml
+++ b/service/application/pom.xml
@@ -17,7 +17,6 @@
     <properties>
         <springdoc.version>1.6.11</springdoc.version>
         <snakeyaml.version>2.0</snakeyaml.version>
-        <spring-cloud-starter-sleuth.version>3.1.5</spring-cloud-starter-sleuth.version>
         <spring-security-rsa.version>1.1.1</spring-security-rsa.version>
         <spring-kafka.version>3.0.10</spring-kafka.version>
         <kafka-clients.version>3.5.0</kafka-clients.version>
@@ -61,11 +60,6 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-data-jpa</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.springframework.cloud</groupId>
-            <artifactId>spring-cloud-starter-sleuth</artifactId>
-            <version>${spring-cloud-starter-sleuth.version}</version>
         </dependency>
         <dependency>
             <groupId>org.springframework.security</groupId>
@@ -253,6 +247,11 @@
         <dependency>
             <groupId>io.micrometer</groupId>
             <artifactId>micrometer-observation</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-observation-test</artifactId>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
--- a/service/application/pom.xml
+++ b/service/application/pom.xml
@@ -240,6 +240,25 @@
             <version>1.1.2-SNAPSHOT</version>
         </dependency>
 
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-tracing-bridge-brave</artifactId>
+            <!--version>1.1.1</version-->
+        </dependency>
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>context-propagation</artifactId>
+            <version>1.0.3</version>
+        </dependency>
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-observation</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-actuator</artifactId>
+        </dependency>
+
         <!-- for testing -->
         <dependency>
             <groupId>org.hamcrest</groupId>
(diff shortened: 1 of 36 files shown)


service/application/pom.xml
```diff
@@ -16,7 +17,6 @@
     <properties>
         <springdoc.version>1.6.11</springdoc.version>
         <snakeyaml.version>2.0</snakeyaml.version>
-        <spring-cloud-starter-sleuth.version>3.1.5</spring-cloud-starter-sleuth.version>
```
gregmeldrum, line 19: Removed sleuth since it is not supported in Spring Boot 3

service/application/pom.xml
```diff
@@ -239,6 +234,30 @@
             <version>1.1.2-SNAPSHOT</version>
         </dependency>
 
+        <dependency>
```
gregmeldrum, line 237: Added micrometer as the new tracing library (instead of sleuth)

service/application/src/main/java/com/solace/maas/ep/common/messages/ScanDataImportMessage.java
```diff
@@ -16,8 +16,6 @@ public class ScanDataImportMessage extends MOPMessage {
 
     String scanId;
 
-    String traceId;
```
gregmeldrum, line 19: The MOPMessage class already has a traceId field so no need to introduce a new field (which eclipses the existing one).

This is done for all of the message type defintions.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/logging/StreamingAppender.java
```diff
@@ -2,11 +2,11 @@
 
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.AppenderBase;
-import com.solace.maas.ep.event.management.agent.repository.model.route.RouteEntity;
```
gregmeldrum, line 5: The "Save Actions" plugins sometimes mess with the import field order.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/scanManager/rest/DataImportControllerImpl.java
```diff
@@ -34,17 +34,20 @@ public class DataImportControllerImpl implements DataImportController {
 
     private final ImportService importService;
 
+    private final Tracer trace;
```
gregmeldrum, line 37: Added a tracer in order to get the traceId (which is done at line 50). The MDC.get was not working, so I got it from the current tracer context instead.

service/application/src/main/resources/logback-spring.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
 
-    <property name="LOG_LEVEL_PATTERN" value="%clr(%5p) %clr([%X{X-B3-TraceId:-}]){yellow}"/>
+    <property name="LOG_LEVEL_PATTERN" value="%clr(%5p) %clr([%X{traceId:-}]){yellow}"/>
```
gregmeldrum, line 4: Micrometer uses `traceId` instead of `X-B3-TraceId`.

service/plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/constants/RouteConstants.java
```diff
@@ -9,7 +9,9 @@ public class RouteConstants {
 
     public static final String SCAN_ID = "SCAN_ID";
 
-    public static final String TRACE_ID = "TRACE_ID";
+    public static final String TRACE_ID = "traceId";
```
gregmeldrum, line 12: Changed "TRACE_ID" to "traceId" so that the logger will automatically pick it up when using this as the MDC context key.

service/plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/mop/MOPMessage.java
```diff
@@ -74,6 +74,9 @@ public enum DestinationType {
     private String destinationName;
     private DestinationType destinationType;
 
+    private String actorId;
```
gregmeldrum, line 77: Adding actorId and orgId to align with maas-base.

slunelsolace (approved):

PR #128, opened 2023-10-27
update the conditions to create beans based on eventPortal section

AHabes:

Update the conditions to create the spring beans based on the existence of the `eventPortal` section in the `application.yml` file

By providing default values for the `EventPortalProperties` when the `eventPortal` doesn't exist in the `application.yml` file. Also, updating the conditions to instantiate classes based on the value of `standalone` flag.

Manually by starting the EMA using different configs in `application.yml` file, e.g., with and without `eventPortal` section, then running scans in online and offline modes

N/A

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/config/eventPortal/EventPortalProperties.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/config/eventPortal/EventPortalProperties.java
@@ -5,16 +5,19 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.PropertySource;
 
+import java.util.List;
+
 @Data
 @Configuration
 @PropertySource("classpath:application.yml")
 @ConfigurationProperties(prefix = "event-portal")
 public class EventPortalProperties {
-    private String runtimeAgentId;
+    private String runtimeAgentId = "standalone";
 
-    private String organizationId;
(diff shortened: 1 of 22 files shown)


service/application/src/main/java/com/solace/maas/ep/event/management/agent/config/eventPortal/EventPortalProperties.java
```diff
@@ -5,16 +5,19 @@
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.PropertySource;
 
+import java.util.List;
```
gregmeldrum, line 8: I don't think you need the changes in this file. The ConfigurationProperties are applied after the ConditionalOnProperty, so it won't make a difference.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/publisher/HeartbeatGenerator.java
```diff
@@ -14,7 +15,8 @@
 
 @ExcludeFromJacocoGeneratedReport
 @Component
-@ConditionalOnExpression("${eventPortal.gateway.messaging.standalone} == false and ${eventPortal.gateway.messaging.enableHeartbeats}")
+@ConditionalOnExpression("${eventPortal.gateway.messaging.enableHeartbeats}")
```
gregmeldrum, line 18: Do you need to turn this into a ConditionalOnProperty as well?

gregmeldrum (approved):

PR #132, opened 2023-11-23
DATAGO-64296: initial commit for terraform plugin

gregmeldrum:
This PR adds the new terraform plugin to the EMA. The plugin is added as a new module.
Additionally, a temporary REST API has been added to allow testing
The REST API follows the newly approved command model

The terraformClient has been copied and modified from https://github.com/microsoft/terraform-spring-boot
The RESP API is added to the `application` module since this module takes care of the northbound interface
The terraform specific code is added to the `terraform-plugin` module.

Some work has been deferred to later stories such as:
* logging support (some logging support is included but it will be enhanced in the future)
* building the docker image containing the terraform executable and solace provider
* adding a messaging API for Solace Cloud Event Portal
* functional testing for all supported usecases (in this PR only simple adding/deleting of queues is tested)
* negative functional testing
* non-functional testing (performance and load)

* Simple functional tests were run using the REST API via POSTMAN commands.
* Integration tests were run using the `TerraformClientRealTests` test class in the terraform-plugin
* Unit tests were added to test the basic terraform command handling using a mocked terraform client.

Just re-iterating, some of the code was taken from open-source (MIT license) at https://github.com/microsoft/terraform-spring-boot
The following classes were copied and modified:
ProcessLauncher
TerraformClient
Our lawyer gave the thumbs up to use this code as long as we include a link to it's origin and a copy of the license under which it was released.


service/plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/mop/MOPProtocol.java
```diff
@@ -3,7 +3,9 @@
 public enum MOPProtocol {
     scanData(2850),
     scanDataControl(2851),
-    EMAHeartbeat(2852);
+    EMAHeartbeat(2852),
+    commandProtocol(2853);
```
gregmeldrum, line 7: This was added for future considerations when we add the messaging API

service/pom.xml
```diff
@@ -237,22 +239,6 @@
                                 </resources>
                             </configuration>
                         </execution>
-                        <execution>
```
gregmeldrum, line 240: Unused execution

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/client/TerraformClient.java
```diff
@@ -0,0 +1,197 @@
+package com.solace.maas.ep.event.management.agent.plugin.terraform.client;
```
gregmeldrum, line 1: This file comes from open source with moderate modifications.

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/configuration/TerraformProperties.java
```diff
@@ -0,0 +1,12 @@
+package com.solace.maas.ep.event.management.agent.plugin.terraform.configuration;
+
+import lombok.Data;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+
+@Service
+@Data
+public class TerraformProperties {
+    @Value("${plugins.terraform.workingDirectoryRoot:/${HOME}/tfconfig}")
```
gregmeldrum, line 10: This is an unorthodox way of setting properties. @Value is not the recommended way to populate config. In this case I wanted a convenient way to set a default value for this value without requiring a change to the application.yml file. Alternatively, we could add a backend story to EP to create a new `plugins.terraform.workingDirectoryRoot` property to the generated application.ylm file.

service/application/src/main/java/com/solace/maas/ep/common/messages/CommandMessage.java
```diff
@@ -0,0 +1,43 @@
...
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+public class CommandMessage extends MOPMessage {
+
+    private String correlationId;
+    private String context;
+    private String messagingServiceId;
```
moodiRealist, line 17: Can we have this named as `serviceId` or some other generic name so it becomes more understandable when we use the command for schemaRegistries?

gregmeldrum: Right, good point!

service/plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/command/model/CommandBundle.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.NoArgsConstructor;
+
+import java.util.List;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class CommandBundle {
+    private String executionType; // (serial / parallel)
```
moodiRealist, line 15: does it make sense for this to be `enum` ?

service/terraform-plugin/pom.xml
```diff
@@ -0,0 +1,160 @@
...
+    <packaging>jar</packaging>
+    <name>Solace Event Management Agent - Terraform Plugin</name>
+    <description>Solace Event Management Agent - Terraform Plugin</description>
+    <properties>
+        <maven.compiler.source>17</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
+        <spring-boot-starter-web.version>3.1.5</spring-boot-starter-web.version>
+        <spring-boot-starter-test.version>3.1.5</spring-boot-starter-test.version>
+        <jupiter.version>5.3.1</jupiter.version>
+        <camel.version>4.0.0-RC2</camel.version>
```
moodiRealist, line 17: Do we need camel on this plugin?

service/terraform-plugin/pom.xml
```diff
@@ -0,0 +1,160 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
+            <version>${commons-collections4.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.solace.maas</groupId>
+            <artifactId>plugin</artifactId>
+            <version>1.3.1-SNAPSHOT</version>
```
moodiRealist, line 61: wasn't this defined in the parent pom file?

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/client/TerraformClient.java
```diff
@@ -0,0 +1,197 @@
...
+
+    private CompletableFuture<Boolean> run(Map<String, String> envVars, String... commands) throws IOException {
+        assert commands.length > 0;
+        ProcessLauncher[] launchers = new ProcessLauncher[commands.length];
+        for (int i = 0; i < commands.length; i++) {
+            launchers[i] = getTerraformLauncher(commands[i], envVars);
+        }
+
+        CompletableFuture<Integer> result = launchers[0].launch().thenApply(c -> c == 0 ? 1 : -1);
+        for (int i = 1; i < commands.length; i++) {
```
moodiRealist, line 151: it reads better if we use `launchers.length` in here even though it is the same value as `commands.length`. Other approach is to define a localVariabe for this length and use it in this method.

Apologies if this is not your modification code and the open-source code.

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/manager/TerraformLogProcessingService.java
```diff
@@ -0,0 +1,168 @@
...
+
+        Path out = Files.createTempFile(logPath, System.currentTimeMillis() + "-"
+                + request.getCorrelationId() + "-job", ".log");
+
+        Files.write(out, logs, Charset.defaultCharset());
+    }
+
+    public CommandResult buildTfCommandResult(List<String> jsonLogs) {
+        if (CollectionUtils.isEmpty(jsonLogs)) {
+            throw new IllegalArgumentException("Cannot extract information empty logs");
```
moodiRealist, line 63: did you mean to say `Cannot extract information from empty logs` ?

moodiRealist, line 63: In case there are no logs, would we return a response back to EP ? or does this exception prevents any response being built and sent back to EP?

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/manager/TerraformLogProcessingService.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+
+    private String extractValueFromLog(Map<String, Object> map, String key) {
+        return String.valueOf(map.get(key));
+    }
+
+    private Map<String, Object> simplifyApplyCompleteLog(Map<String, Object> expandedLogMessage) {
+        if (!isApplyCompleteLog(expandedLogMessage)) {
+            throw new UnsupportedOperationException(
```
moodiRealist, line 101: Seems like we lose this information on the caller:

Screenshot 2023-11-24 at 9 10 09 AM

So you can throw a generic exception in here instead.

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/manager/TerraformLogProcessingService.java
```diff
@@ -0,0 +1,168 @@
...
+                KEY_DIAGNOSTIC_DETAIL, extractDiagnosticDetailMessage(expandedLogMessage.get(KEY_DIAGNOSTIC))
+        );
+
+    }
+
+    private String extractDiagnosticDetailMessage(Object diagnostic) {
+        if (diagnostic == null) {
+            return "";
+        }
+        if (!(diagnostic instanceof Map)) {
```
moodiRealist, line 131: Can you use the new java 17 power here? https://mydeveloperplanet.com/2021/09/28/whats-new-between-java-11-and-java-17/

Screenshot 2023-11-24 at 9 15 01 AM

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/manager/TerraformLogProcessingService.java
gregmeldrum: Yes, the error handling needs some re-work. I'll fix it up.

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/manager/TerraformLogProcessingService.java
gregmeldrum: Yes, I'll fix it.

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/manager/TerraformManager.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    public void execute(CommandRequest request, Command command, Map<String, String> envVars) {
+
+        MDC.put(COMMAND_CORRELATION_ID, request.getCorrelationId());
+        MDC.put(MESSAGING_SERVICE_ID, request.getServiceId());
+        String traceId = MDC.get("traceId");
+        String spanId = MDC.get("spanId");
+
+        log.debug("Executing command {} for ms {} correlationId {} context {}", command.getCommand(), request.getServiceId(),
```
moodiRealist, line 52: please change `ms` to some generic name like `service` since it won't work for schemaRegsitry configPush scenarios

service/terraform-plugin/src/test/java/com/solace/maas/ep/event/management/agent/plugin/terraform/TerraformCommandIT.java
```diff
@@ -0,0 +1,357 @@
...
+        doAnswer(invocation -> {
+            Object arg0 = invocation.getArgument(0);
+            Consumer<String> listener = (Consumer<String>) arg0;
+            logs.forEach(listener);
+            return null;
+        }).when(terraformClient).setOutputListener(any());
+    }
+
+
+    public static String getResourceAsString(Resource resource) {
```
moodiRealist, line 340: Does this method and the one below need to be `public static` or would `private` suffice?

gregmeldrum: Private will do. Updating...

moodiRealist (approved):

PR #136, opened 2023-12-05
DATAGO-65659: Add terraform assets to the docker build

gregmeldrum:
Add the terraform assets to docker since we do not want the user to have to do any setup themselves.

Added the terraform executable and solace provider to the docker directory and the Dockerfile adds them to the image and sets up the path

Tested in docker on Windows, Linux and MacOS. Manually sent a mop message to trigger queue/subscription creation, update and deletion.

No

### DIFF
--- a/service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/configuration/TerraformProperties.java
+++ b/service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/configuration/TerraformProperties.java
@@ -7,6 +7,6 @@ import org.springframework.stereotype.Service;
 @Service
 @Data
 public class TerraformProperties {
-    @Value("${plugins.terraform.workingDirectoryRoot:/${HOME}/tfconfig}")
+    @Value("${plugins.terraform.workingDirectoryRoot:${HOME}/tfconfig}")
     private String workingDirectoryRoot;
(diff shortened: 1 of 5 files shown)


service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/configuration/TerraformProperties.java
```diff
@@ -7,6 +7,6 @@
 @Service
 @Data
 public class TerraformProperties {
-    @Value("${plugins.terraform.workingDirectoryRoot:/${HOME}/tfconfig}")
```
gregmeldrum, line 10: Remove the unneeded leading `/` character.

moodiRealist (approved):

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/configuration/TerraformProperties.java
```diff
@@ -7,6 +7,6 @@
...
+    @Value("${plugins.terraform.workingDirectoryRoot:${HOME}/tfconfig}")
     private String workingDirectoryRoot;
 }
```
CameronRushton, line 12: test comment; please ignore

PR #141, opened 2023-12-12
[DATAGO-64298] rename correlationId to commandCorrelationId in missed spots

moodiRealist:
To rename an important attribute which without this PR, the proper end-to-end flow of sending TF logs to EP won't work.

Renamed correlationId -> commandCorrelationId in code

Manually and updated ITs

Nope


gregmeldrum (approved):
Looks good!

PR #142, opened 2023-12-13
DATAGO-67370: Add a thread pool to command manager

gregmeldrum:
The EMA needs to be able to handle comands in parallel

Added a configurable thread pool to the CommandManager. The default queue size for the thread pool is 1000 which seems more than generous for a single org. Once we move to the cloud we may want to increase this value.

Manually tests in the IDE as well as with docker.

No


service/application/src/main/java/com/solace/maas/ep/event/management/agent/config/eventPortal/EventPortalProperties.java
```diff
@@ -18,6 +18,10 @@ public class EventPortalProperties {
 
     private String topicPrefix;
```
gregmeldrum, line 20: Some initial defaults, we can tweak based on real world usage.

moodiRealist (approved):

PR #144, opened 2023-12-14
DATAGO-64298: rotate command log files.

moodiRealist:
To rotate command log files using logback functionality.

updated logback-spring.xml and removed previous manual logging functionality.

Ran the code multiple times to see logs being rotated and eventually removed:
Screenshot 2023-12-14 at 2 29 15 PM

Also added a new attribute `eventManagementId` to commandLogs message to satisfy a requirement

### DIFF
--- a/service/terraform-plugin/pom.xml
+++ b/service/terraform-plugin/pom.xml
@@ -88,11 +88,6 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <dependency>
-            <groupId>com.google.code.gson</groupId>
-            <artifactId>gson</artifactId>
-            <version>2.10.1</version>
-        </dependency>
     </dependencies>
     <build>
         <plugins>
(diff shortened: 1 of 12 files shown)


service/terraform-plugin/pom.xml
```diff
@@ -88,6 +88,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.10.1</version>
+        </dependency>
     </dependencies>
```
rudraneel-chakraborty, line 96: ☁️ I was wondering if adding another json processing library is needed. I think we already have jackson library available to ema, can we reuse that library?

gregmeldrum (approved):
Looks good!

PR #155, opened 2024-01-23
DATAGO-64061: cli upload functionality

gregmeldrum:

    ...

    ...

    ...

    ...


service/application/src/main/java/com/solace/maas/ep/event/management/agent/cli/CommandLineCommon.java
```diff
@@ -0,0 +1,91 @@
+package com.solace.maas.ep.event.management.agent.cli;
```
gregmeldrum, line 1: Pulled out some of the common code to this file that the scan and upload commands can utilize.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/cli/CommandLineScan.java
```diff
@@ -0,0 +1,107 @@
+package com.solace.maas.ep.event.management.agent.cli;
+
```
gregmeldrum, line 2: This is existing functionality that has been refactored into it's own class.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/cli/MultipartFileWrapper.java
```diff
@@ -0,0 +1,61 @@
+package com.solace.maas.ep.event.management.agent.cli;
```
gregmeldrum, line 1: The existing file import functionality requires an input object of type MultipartFile that is used to get the InputStream for the scan file. This wrapper class was created to do just that and nothing else.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/cli/EmaCommandLine.java
```diff
@@ -1,212 +1,85 @@
...
-                throw new RuntimeException("Scan is stuck in INITIATED state. Check broker URL and credentials.");
+        try {
+            if (args.length >= 1) {
+
+                String type = args[0];
+
+                if ("scan".equals(type)) {
+                    if (args.length > 2) {
+                        String messagingServiceId = args[1];
+                        String filePathAndName = args[2];
```
moodiRealist, line 47: In the case that there are more than 3 args provided, we are ignoring the other args. Is that a better practice than throwing a too many args error?

gregmeldrum: Yes, good idea, I'll add a check for too many args.

PR #161, opened 2024-02-09
status set on CommandMessage only right before sending mop response

CameronRushton:

    ...

    ...

    ...

    ...

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/CommandManager.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/CommandManager.java
@@ -63,7 +63,7 @@ public class CommandManager {
                     log.error("Error running command", e);
                     Command firstCommand = request.getCommandBundles().get(0).getCommands().get(0);
                     setCommandError(firstCommand, (Exception) e);
-                    sendResponse(request);
+                    finalizeAndSendResponse(request);
                     return null;
                 });
     }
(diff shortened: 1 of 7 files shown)


service/plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/command/model/Command.java
```diff
@@ -19,4 +19,11 @@ public class Command {
     private Map<String, String> parameters;
     private Boolean ignoreResult;
     private CommandResult result;
+
+    public boolean hasSignificantErrorResult() {
+        if (!Boolean.TRUE.equals(ignoreResult) && result != null) {
+            return JobStatus.error.equals(result.getStatus());
+        }
```
CameronRushton, line 26: This conditional works when ignoreResult is null, so null is treated as false.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/CommandManager.java
```diff
@@ -105,10 +105,11 @@ public void configPush(CommandMessage request) {
                 }
             }
         }
-        sendResponse(request);
+        finalizeAndSendResponse(request);
     }
 
-    private void sendResponse(CommandMessage request) {
+    private void finalizeAndSendResponse(CommandMessage request) {
+        request.determineStatus();
```
CameronRushton, line 112: This method is on the object because I think it has better encapsulation since this entire operation takes place within itself.

PR #163, opened 2024-02-09
DATAGO-70002: Import support in EMA

gregmeldrum:

Introduces v2 of the EMA command interface which allows the agent to rebuild state after a restart.

EMA now supports:
* the tofu "import" command
* logs are not sent to ep-core during the import command
* environment variables in the command parameters
* a top level result for all commands
* true support for exitOnFailure
* true support for ignore result
 

Manual testing using Postman to send MOP messages

No


service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/manager/TerraformManager.java
```diff
@@ -52,65 +47,78 @@ public void execute(CommandRequest request, Command command, Map<String, String>
...
-    }
+        List<String> logOutput = TerraformUtils.setupTerraformClient(terraformClient, configPath, logToConsole);
 
-    private static String executeTerraformCommand(Command command, Map<String, String> envVars, Path configPath, TerraformClient terraformClient) throws IOException, InterruptedException, ExecutionException {
-        String commandVerb = command.getCommand();
         switch (commandVerb) {
+            case "import" -> importCommand(envVars, configPath, terraformClient, logOutput);
             case "apply" -> {
-                writeHclToFile(command, configPath);
-                Boolean planSuccessful = terraformClient.plan(envVars).get();
```
gregmeldrum, line 105: Calling terraform `plan` before each `apply` was not adding value and only adds to the amount of time it takes to execute and doubles the number of logs. Removed it streamline the apply.

service/application/src/main/java/com/solace/maas/ep/common/messages/CommandMessage.java
```diff
@@ -33,9 +38,19 @@ public CommandMessage(String serviceId,
...
     }
 
+    public void determineStatus() {
+        boolean hasAtLeastOneError = false;
+        if (CollectionUtils.isNotEmpty(commandBundles)) {
+            hasAtLeastOneError = commandBundles.stream().anyMatch(bundle ->
+                    bundle.getCommands().stream().anyMatch(Command::hasSignificantErrorResult));
+        }
+        this.setStatus(hasAtLeastOneError ? JobStatus.error : JobStatus.success);
+    }
```
CameronRushton, line 52: Noting what I/we were thinking here like a mini ADR:
It would be ideal if this logic was on a BO and both the BO and mop message had a `status` field. Since we start with this command message and send the same thing back as a response, there's not a clear separation. Since the places we set command errors are in the application module _and_ tf plugin module that uses the BO (CommandRequest), it's less error prone & a bit cleaner to calculate the status just before we send the response.

This pojo is doing the calculation because it requires no outside data going in or out for better encapsulation.
We don't set the status first, then calculate before sending to avoid a case where we report the default status set in the beginning. If we don't set the status first and there's a bug (we don't calculate the status), we'll get null and can easily notice the issue.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/CommandManager.java
```diff
@@ -103,12 +103,18 @@ public void configPush(CommandMessage request) {
                     log.error("Error executing command", e);
                     setCommandError(command, e);
                 }
+                if (bundle.getExitOnFailure() && command.getResult().getStatus() == JobStatus.error
+                        && !Boolean.TRUE.equals(command.getIgnoreResult())) {
+                    break;
+                }
```
CameronRushton, line 109: sonarlint really wants all Boolean boxed values to be unboxed safer accounting for a null value like the `Boolean.TRUE.equals()` here.
I'd like to at least change this around so that it reads `bundle.getExitOnFailure() &&  JobStatus.error.equals(command.getResult().getStatus())`. Sonarlint is happy with `(Boolean.TRUE.equals(bundle.getExitOnFailure() &&  JobStatus.error.equals(command.getResult().getStatus()))` but it seems too verbose to me.

Also, I thought getResult() could be null if `ignoreResult` is true, yeah? This would be typical for import actions. The error result would be passed on to the apply. Looks like the logic here is always setting the result on the command that failed. I don't think this was mentioned in the ADR so I want to add what we decide to it.

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/manager/TerraformManager.java
```diff
@@ -52,65 +47,78 @@ public void execute(CommandRequest request, Command command, Map<String, String>
...
-        terraformClient.setOutputListener(tfLog -> {
-            output.add(tfLog);
-            logToConsole(tfLog);
-        });
-        return output;
+    private static void setEnvVarsFromParameters(Command command, Map<String, String> envVars) {
+        if (command.getParameters() != null && command.getParameters().containsKey("environment") &&
+                command.getParameters().get("environment") instanceof Map) {
+            Map<String, String> environmentMap = (Map<String, String>) command.getParameters().get("environment");
+            environmentMap.keySet().forEach(key -> envVars.put("TF_VAR_" + key, environmentMap.get(key)));
```
CameronRushton, line 71: What env are these variables to be set in? I was thinking it was the EMA's env but it looks like it's transformed here to be exported as an environment variable later in a different context.

service/application/src/main/java/com/solace/maas/ep/common/messages/CommandMessage.java
gregmeldrum: Good suggestion, I like the idea of adding a BO layer for the reasons you mention. I'll add it in.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/CommandManager.java
gregmeldrum: Right, good point about the boolean logic/null checking. I'll re-work this logic.
I'll look at the terraform plugin logic to make sure we're not setting the result if the `ignoreResult` flag is set.

service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/manager/TerraformManager.java
gregmeldrum: The `tofu` commands are run on a process that is created by the `TerraformClient`. These environment variables need to be set on the process that is created, so `tofu` can do the appropriate variable substitution. If you look at `TerraformClient::getTerraformLauncher`, this is the method that sets up the process launcher and sets the environment variables for the process.

CameronRushton (approved):
No need to implement a BO layer/do any drastic refactoring for this PR. This is totally mergable. Thanks!

service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/CommandManager.java
```diff
@@ -118,14 +122,22 @@ private static boolean exitEarlyOnFailedCommand(CommandBundle bundle, Command co
...
         );
-        commandPublisher.sendCommandResponse(request, topicVars);
+        CommandMessage response = new CommandMessage(request.getServiceId(),
+                request.getCommandCorrelationId(),
+                request.getContext(),
+                request.getStatus(),
+                request.getCommandBundles());
+        response.setOrgId(eventPortalProperties.getOrganizationId());
+        response.setTraceId(MDC.get(TRACE_ID));
+        response.setActorId(MDC.get(ACTOR_ID));
```
CameronRushton, line 139: We should try doing this in the mapper.
We'll probably have to add this to the no args constructor because that's what mapstruct will use by default (it will actually use whatever constructor has a `@Default` annotation if available - even if we make our own annotation).
```
withMessageType(MOPMessageType.generic)
                .withProtocol(MOPProtocol.epConfigPush)
                .withVersion("1")
                .withUhFlag(MOPUHFlag.ignore);
``` 

Then maybe have an after mapping for the traceId and actorId... and maybe even for the orgId...
```
@AfterMapping
    protected void assignMopMessageFields(@MappingTarget CommandMessage target, CommandRequest source) {
        target.setTraceId(MDC.get(TRACE_ID));
        // etc...
    }
```

PR #167, opened 2024-02-28
DATAGO-72210: Fix for subscription aggregation bug

gregmeldrum:

The subscription scan was ending prematurely after getting a few queues without subscriptions. It was erroneously signaling the end of the subscription scan.

The subscription scan should only conclude once the last batch of data has been processed. Therefore, the condition to end the scan when it's empty has been enhanced to verify whether it's handling the final batch of data.

It was manually tested with the standard kafka and Solace configurations. I also tested with an empty schema registry to make sure the base case of 0 entities is also working. Also, I created a scenario similar to Mars, where I have a big batch of queues with no subscriptions. Without the fix, the scan failed, and with the fix, it succeeded.

Aly Rasmy tested this against the customer data, and the EMA scan appears to have completed correctly from the logs, but we ran into a timeout issue which will be address by increasing the ep-core timeout.

### DIFF
--- a/service/plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/route/handler/base/DataAggregationRouteBuilder.java
+++ b/service/plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/route/handler/base/DataAggregationRouteBuilder.java
@@ -14,9 +14,6 @@ import org.apache.camel.model.dataformat.JsonLibrary;
 
 import java.util.Objects;
 
-import static com.solace.maas.ep.event.management.agent.plugin.constants.HeaderConstants.DATA_PROCESSING_COMPLETE;
-import static com.solace.maas.ep.event.management.agent.plugin.constants.HeaderConstants.DESTINATIONS;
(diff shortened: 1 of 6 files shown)


service/plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/route/handler/base/DataAggregationRouteBuilder.java
```diff
@@ -86,7 +87,8 @@ public void configure() throws Exception {
                 .log(LoggingLevel.TRACE, "agg complete ${body}")
                 .process(processor)
                 // Checking for empty scan types.
-                .choice().when(simple("${body.size} == 0"))
+                .choice().when(and(simple("${body.size} == 0"),
+                        header(RouteConstants.DATA_PROCESSING_COMPLETE).isEqualTo(true)))
```
gregmeldrum, line 91: This is the main change, the rest are just to improve the code. Here I added an and clause that checks that we only mark the scan types as complete when we are processing the final batch of data (DATA_PROCESSING_COMPLETE is set to true).

PR #169, opened 2024-02-29
DATAGO-72530: Prisma exclusion list checker in EMA should use correct format for package identifiers

shooshmand-sol:
The format for checking excluded libraries for Prisma vulenrabilites, should be `library:version` instead of `library-version` to match the database schema which stores list of excluded libraries.

Simple change to use `library:version` when checking Prisma exclusion list


gregmeldrum (approved):

johnvincentcorpuz (approved):
👍

PR #170, opened 2024-03-04
DATAGO-72325: Add OS independent home directory and file separator character

gregmeldrum:

By default, the EMA tries to create a directory using linux/mac friendly path syntax. This causes a bug when running on Windows

Change the linux/mac friendly path syntax with a syntax that works on linux, mac and Windows.

Tested on a Windows VM using UTM. The EMA started up and was able to run a scan.

No

### DIFF
--- a/service/terraform-plugin/src/test/java/com/solace/maas/ep/event/management/agent/plugin/terraform/TerraformCommandIT.java
+++ b/service/terraform-plugin/src/test/java/com/solace/maas/ep/event/management/agent/plugin/terraform/TerraformCommandIT.java
@@ -12,7 +12,9 @@ import com.solace.maas.ep.event.management.agent.plugin.terraform.configuration.
 import com.solace.maas.ep.event.management.agent.plugin.terraform.manager.TerraformManager;
 import org.apache.commons.lang.StringUtils;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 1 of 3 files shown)


service/terraform-plugin/src/test/java/com/solace/maas/ep/event/management/agent/plugin/terraform/TerraformCommandIT.java
```diff
@@ -209,7 +217,7 @@ public void testImportResourceWithSomeResourcesExistingAndSomeNot() throws IOExc
                 "\tid = \"default/MyConsumer1/a%2Fb%2Fc\"\n" +
                 "}";
 
-        String content = Files.readString(Path.of(terraformProperties.getWorkingDirectoryRoot() + "/app123-ms1234/import.tf"));
+        String content = Files.readString(Path.of(terraformProperties.getWorkingDirectoryRoot() + "/app123-ms1234/sync.tf"));
```
gregmeldrum, line 220: This test was passing for me locally since I still had the import.tf file in my local directory but it should have failed since we renamed the file to `sync.tf`. I changed the workingDirectoryRoot for tests to use `java.io.tmpdir` instead of `${HOME}/tfcommands` so it wouldn't just use the same directory every time the tests are run.

rudraneel-chakraborty (approved):

PR #171, opened 2024-03-19
DATAGO-72931: Identical scanIds bug on Windows

mynecker:
SecureRandom generates a sequence of identical numbers with a static seed on Windows, which leads to identical scandIds. This implementation fixes this issue. 

Removing a static seed and utilizing default seed generation of SecureRandom, if `orgId` is not defined. 

Unit tests executed on MacOS on Intel and unit tests executed on Windows. 

Ideally cross-check on another Windows box the fixed agent.


gregmeldrum (approved):
Look good!

PR #172, opened 2024-03-21
DATAGO-73412: event-management-agent:spring-web:6.0.17,Type: app_dependency,Prisma Cloud Vulnerability

CameronRushton:

Uprev spring-web from 6.0.17 -> 6.0.18 to solve a vulnerability.

Adding an explicit dependency to spring-web v6.0.18 in each module's pom. It's best to do it this way because upgrading spring parent isn't going to get us a good version that isn't vulnerable. We must override it. We also can't simply add the version number to the properties section because it's not a variable to be overridden like that on spring's side. So, we explicitly add the dependency override until we can upgrade to a version of spring-web that is safe in spring-parent 3.1.10 or 3.2.4 which don't exist yet. 

Ran the EMA and saw no issues.
See attached dependencies before and after from `mvn dependency:tree`.

AFTER
[dependencies_AFTER.txt](https://github.com/SolaceProducts/event-management-agent/files/14711833/dependencies_AFTER.txt)

BEFORE
[dependencies_BEFORE.txt](https://github.com/SolaceProducts/event-management-agent/files/14711834/dependencies_BEFORE.txt)

### DIFF
--- a/service/confluent-schema-registry-plugin/pom.xml
+++ b/service/confluent-schema-registry-plugin/pom.xml
@@ -17,6 +17,7 @@
         <maven-shade-plugin.version>3.3.0</maven-shade-plugin.version>
         <jackson-dataformat-cbor.version>2.13.4</jackson-dataformat-cbor.version>
         <maven-surefire-plugin.version>2.22.0</maven-surefire-plugin.version>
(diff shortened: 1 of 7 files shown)


gregmeldrum (approved):
Thanks Cameron!

PR #173, opened 2024-04-19
DATAGO-74884: event-management-agent:spring-web-6.0.18.jar, HIGH, 8.1, vulnerability detected

CameronRushton:

Uprev to spring 3.1.11 to fix spring-web vulnerability

Upreving the version on all plugins and the main app, removing the spring-web overridden dependency, and upreving the main spring version which has the latest spring-web.

I was able to clean+compile and run the EMA locally. Didn't run data through it though. 

We need to create a release for the docker image & if this isn't fixed, it's going to block the rc pipeline next week.

Updated dependency tree
[dependencies_AFTER.txt](https://github.com/SolaceProducts/event-management-agent/files/15044131/dependencies_AFTER.txt)

### DIFF
--- a/service/application/pom.xml
+++ b/service/application/pom.xml
@@ -4,11 +4,11 @@
     <parent>
         <groupId>com.solace.maas</groupId>
         <artifactId>maas-event-management-agent-parent</artifactId>
-        <version>1.6.4-SNAPSHOT</version>
+        <version>1.6.5-SNAPSHOT</version>
         <relativePath>..</relativePath>
     </parent>
     <artifactId>event-management-agent</artifactId>
-    <version>1.6.4-SNAPSHOT</version>
+    <version>1.6.5-SNAPSHOT</version>
     <packaging>jar</packaging>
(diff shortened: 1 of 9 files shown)


service/application/pom.xml
```diff
@@ -4,11 +4,11 @@
     <parent>
         <groupId>com.solace.maas</groupId>
         <artifactId>maas-event-management-agent-parent</artifactId>
-        <version>1.6.4-SNAPSHOT</version>
+        <version>1.6.5-SNAPSHOT</version>
```
moodiRealist, line 7: I don't think we need any of these 1.6.4 to 1.6.5 changes. As soon as we run the release, it will update all these versions to the requested number, e.g. 1.6.5 or 1.7.0 or 2.0.0. Right @gregmeldrum ?

gregmeldrum: Correct, it should automatically update

moodiRealist (approved):

rudraneel-chakraborty (approved):

PR #179, opened 2024-05-06
DATAGO-74222: Upgrade EMA to spring 3.2.5

CameronRushton:

Upgrading the EMA to spring 3.2.5

Change the spring parent dependencies to 3.2.5 and fixing the beans that are now broken. Some annotations were missing or not working.

I did a config push and audit.
I need to exercise the code more in standalone mode and ensure the affected table columns are exercised.

See comments

### DIFF
--- a/service/application/pom.xml
+++ b/service/application/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>com.solace.maas</groupId>
@@ -13,7 +14,6 @@
     <name>Solace Event Management Agent - Application</name>
     <description>Solace Event Management Agent - Application</description>
     <properties>
-        <springdoc.version>1.6.11</springdoc.version>
         <snakeyaml.version>2.0</snakeyaml.version>
         <spring-security-rsa.version>1.1.1</spring-security-rsa.version>
         <spring-kafka.version>3.0.10</spring-kafka.version>
@@ -88,7 +88,7 @@
         <dependency>
             <groupId>org.springdoc</groupId>
             <artifactId>springdoc-openapi-starter-webmvc-ui</artifactId>
-            <version>2.0.0</version>
+            <version>2.2.0</version>
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
@@ -306,11 +306,6 @@
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok-mapstruct-binding</artifactId>
             <version>0.2.0</version>
-        </dependency>
-        <dependency>
-            <groupId>org.springdoc</groupId>
-            <artifactId>springdoc-openapi-ui</artifactId>
-            <version>${springdoc.version}</version>
         </dependency>
         <dependency>
             <groupId>jakarta.servlet</groupId>
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/CommandLogStreamingProcessor.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/CommandLogStreamingProcessor.java
@@ -12,6 +12,7 @@ import com.solace.maas.ep.event.management.agent.plugin.terraform.manager.Terraf
 import com.solace.maas.ep.event.management.agent.publisher.CommandLogsPublisher;
 import lombok.extern.slf4j.Slf4j;
 import org.slf4j.MDC;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
 
 import java.io.IOException;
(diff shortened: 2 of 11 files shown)


service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/CommandLogStreamingProcessor.java
```diff
@@ -32,6 +33,7 @@
 
 @Component
 @Slf4j
+@ConditionalOnProperty(name = "event-portal.gateway.messaging.standalone", havingValue = "false")
 public class CommandLogStreamingProcessor {
```
CameronRushton, line 37: Only used for config push and the other config push related beans have this conditional. Without it, it tries to find the dependent beans but finds none when not in standalone mode.

service/application/pom.xml
```diff
@@ -84,11 +87,12 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
+            <version>${spring-boot.version}</version>
```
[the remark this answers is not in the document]
CameronRushton: I think it does, but I was seeing a lot of warnings how there wasn't a version and it had to assume. On these ones specifically, I may have added them for no reason. I was in the area and adding them for others so I figured we might as well. I'm going to try removing some and seeing if the warnings return.

service/application/src/test/java/com/solace/maas/ep/event/management/agent/scanManager/rest/DataImportControllerTests.java
```diff
@@ -56,8 +56,6 @@ public void testDataImportControllerRead() {
 
         assertThat(reply.getStatusCode()).isEqualTo(HttpStatus.OK);
         assertThat(reply.getBody()).contains("Manual import started.");
-
-        assertThatNoException();
```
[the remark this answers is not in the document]
CameronRushton: Not for these tests. A few things to note here:
- assertThatNoException() isn't used this way so it was being useless here.
- If we want to test that there's no exception, we should just check the status code because this code is set up to return a 400 on error as seen in other tests as well
- Any other uncaught exceptions will fail the test

service/kafka-plugin/src/test/java/com/solace/maas/ep/event/management/agent/plugin/kafka/route/delegate/KafkaRouteDelegateImplTests.java
```diff
@@ -35,8 +34,7 @@ void testGenerateKafkaTopicListingRouteList() {
                 kafkaRouteDelegate.generateRouteList(destinations, List.of(), KafkaScanType.KAFKA_TOPIC_LISTING.name(),
                 "service1");
 
-        assertThatNoException();
```
CameronRushton, line 38: We dont need this for these tests either because if there's an uncaught exception, it will fail the test and if there is an exception, it returns an empty list which we specifically assert.

service/terraform-plugin/pom.xml
```diff
@@ -94,33 +94,12 @@
             <plugin>
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
+                <version>${spring-boot.version}</version>
                 <configuration>
                     <mainClass>com.solace.maas.ep.event.management.agent.plugin.terraform.TerraformApplication
                     </mainClass>
                 </configuration>
             </plugin>
-            <plugin>
```
CameronRushton, line 102: duplicate

service/solace-plugin/pom.xml
```diff
@@ -105,32 +105,11 @@
             <plugin>
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
+                <version>${spring-boot.version}</version>
                 <configuration>
                     <mainClass>com.solace.maas.ep.event.management.agent.plugin.solace.SolaceApplication</mainClass>
                 </configuration>
             </plugin>
-            <plugin>
```
CameronRushton, line 112: duplicate

service/application/pom.xml
```diff
@@ -334,20 +333,14 @@
                     </excludes>
                 </configuration>
             </plugin>
-            <plugin>
```
CameronRushton, line 337: duplicate

service/application/pom.xml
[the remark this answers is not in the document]
CameronRushton: Turns out, it does get upset when it's not there.
```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for org.springframework.boot:spring-boot-starter-web:jar is missing. @ line 76, column 21
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project com.solace.maas.plugin.terraform:terraform-plugin:1.6.6-SNAPSHOT (/Users/cameronrushton/Code/event-management-agent/service/terraform-plugin/pom.xml) has 1 error
[ERROR]     'dependencies.dependency.version' for org.springframework.boot:spring-boot-starter-web:jar is missing. @ line 76, column 21
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
```

service/kafka-plugin/pom.xml
```diff
@@ -10,12 +11,12 @@
...
+        <spring-kafka.version>3.1.4</spring-kafka.version>
         <kafka-clients.version>3.5.0</kafka-clients.version>
-        <spring-boot-version>3.1.11</spring-boot-version>
-        <commons-collections4.vresion>4.4</commons-collections4.vresion>
-        <jupiter.version>5.3.1</jupiter.version>
-        <camel.version>4.2.0</camel.version>
+        <spring-boot.version>3.2.5</spring-boot.version>
+        <commons-collections4.version>4.4</commons-collections4.version>
+        <jupiter.version>5.10.2</jupiter.version>
+        <camel.version>4.5.0</camel.version>
```
moodiRealist, line 19: Are these upversions needed? If yes, have we run scans against kafka brokers to make sure it still works?

CameronRushton: Yeah, they are necessary. I'll see about testing it.

PR #183, opened 2024-06-07
Uprev from 1.6.7-SNAPSHOT to 1.6.8-SNAPSHOT since 1.6.7 was already released

CameronRushton:

 The release pipeline was triggered incorrectly using the same version for both fields, so 1.6.7 was released but the snapshot version is now 1.6.7 too. It should be 1.6.8-SNAPSHOT then we release 1.6.8.

Manually uprev the version.

I ran the EMA locally and it booted up.

I also edited the config because the path was deprecated.

### DIFF
--- a/service/application/src/main/resources/application.yml
+++ b/service/application/src/main/resources/application.yml
@@ -36,7 +36,7 @@ spring:
     allow-bean-definition-overriding: true
 
 camel:
-  springboot:
+  main:
     use-mdc-logging: true
   errorHandling:
     maximumRedeliveries: 10
--- a/service/application/src/main/resources/application.yml
+++ b/service/application/src/main/resources/application.yml
@@ -36,7 +36,7 @@ spring:
     allow-bean-definition-overriding: true
 
 camel:
-  main:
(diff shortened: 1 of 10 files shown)


moodiRealist (approved):

service/application/src/main/resources/application.yml
```diff
@@ -36,7 +36,7 @@ spring:
     allow-bean-definition-overriding: true
 
 camel:
-  springboot:
+  main:
```
gregmeldrum, line 39: Was this an intentional change?
We would need to test that the MDC attributes are still working for a scan initiated from Solace cloud (we carry state through MDC like the traceId and actorId that needs to be present and eventually sent back to Solace Cloud in the response).

CameronRushton: Yeah, I saw that it's deprecated so I switched it over. I didn't test it thoroughly, so I will remove it and get that sorted out on ep-core's side and EMA's side another time.

gregmeldrum (approved):

PR #184, opened 2024-06-13
DATAGO-75198: Common code for DATAGO-75198 and DATAGO-77217

rudraneel-chakraborty:

These classes will be used for dynamic configuration for configpush and scan.  Creating the PR to unblock development

Java

NA

    ...

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/common/model/ResourceConfigurationType.java
+++ b/service/application/src/main/java/com/solace/maas/ep/common/model/ResourceConfigurationType.java
@@ -1,6 +1,5 @@
 package com.solace.maas.ep.common.model;
 
 public enum ResourceConfigurationType {
-    SOLACE,
-    KAFKA,
+    SOLACE
 }
--- a/service/application/src/main/java/com/solace/maas/ep/common/model/ResourceConfigurationType.java
+++ b/service/application/src/main/java/com/solace/maas/ep/common/model/ResourceConfigurationType.java
@@ -1,0 +1,6 @@
+package com.solace.maas.ep.common.model;
+
+public enum ResourceConfigurationType {
+    SOLACE,
+    KAFKA,
+}
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/service/EventBrokerResourceConfigToEventConverter.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/service/EventBrokerResourceConfigToEventConverter.java
@@ -1,0 +1,82 @@
+package com.solace.maas.ep.event.management.agent.service;
+
+import com.solace.maas.ep.common.model.EventBrokerAuthenticationConfiguration;
+import com.solace.maas.ep.common.model.EventBrokerConnectionConfiguration;
+import com.solace.maas.ep.common.model.EventBrokerCredentialConfiguration;
+import com.solace.maas.ep.common.model.EventBrokerResourceConfiguration;
+import com.solace.maas.ep.event.management.agent.event.MessagingServiceEvent;
+import com.solace.maas.ep.event.management.agent.plugin.messagingService.event.AuthenticationDetailsEvent;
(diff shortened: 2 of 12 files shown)


service/application/src/main/java/com/solace/maas/ep/common/model/ResourceConfigurationType.java
```diff
@@ -0,0 +1,6 @@
+package com.solace.maas.ep.common.model;
+
+public enum ResourceConfigurationType {
+    SOLACE,
+    KAFKA,
```
moodiRealist, line 5: Since we haven't added CONFLUENT_SCHEMA_REGISTRY here, I suggest removing `KAFKA` for the time being until we are ready to support that. Will be an easy update.

moodiRealist (dismissed):
Approved with a small comment.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/subscriber/messageProcessors/DynamicResourceConfigurationHelper.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+        try {
+
+            List<MessagingServiceEvent> solaceEventBrokerResources = resources.stream()
+                    .filter(resource -> resource.getResourceConfigurationType() == ResourceConfigurationType.SOLACE)
+                    .map(eventBrokerResourceConfigToEventConverter::mapToMessagingServiceEvent)
+                    .toList();
+            messagingServiceDelegateService.addMessagingServices(solaceEventBrokerResources)
+                    .forEach(messagingServiceEntity ->
+                            log.info("Loaded [{}] resource with id: [{}] and name: [{}] from message payload.",
```
moodiRealist, line 40: I think this logLevel can be lowered to `debug`. Also, since this method can be called for a list of 10 brokers, I think it is better for this part to produce 1 log instead of 10.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/service/EventBrokerResourceConfigToEventConverter.java
```diff
@@ -0,0 +1,82 @@
...
+
+        ConnectionDetailsEvent connectionDetailsEvent = new ConnectionDetailsEvent();
+        EventBrokerConnectionConfiguration connectionDetail = eventBrokerResource.getConnections().get(0);
+
+        connectionDetailsEvent.setName(eventBrokerResource.getName());
+        connectionDetailsEvent.setUrl(connectionDetail.getUrl());
+        connectionDetailsEvent.setMessagingServiceId(eventBrokerResource.getId());
+        connectionDetailsEvent.setProperties(List.of(
+                EventProperty.builder()
+                        .name("sempPageSize")
```
moodiRealist, line 36: I think it might help readability if this class or method name contains `SolaceBroker` instead of the generic names.

moodiRealist (commented):
Made some comments that worth addressing IMO.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/service/EventBrokerResourceConfigToEventConverter.java
rudraneel-chakraborty: Done

service/application/src/main/java/com/solace/maas/ep/event/management/agent/subscriber/messageProcessors/DynamicResourceConfigurationHelper.java
rudraneel-chakraborty: Changed log level to debug. 
Additionally, I don't think we can have more than 1 dynamic broker config (Config push is done against a single broker, so is scan) Moreover, it may make sense to see what broker configurations were added successfully and what broker config was faulty and failed

moodiRealist (approved):

PR #189, opened 2024-06-27
DATAGO-78232: Add new spring variable to override the use of user.home

gregmeldrum:

When running the docker image in k8s, the user.home get's set to '?' which causes the EMA to fail to load

Anywhere that we use user.home, we add another variable that can be set on the command line (or via an env var) that overrides the use of user.home, so the "command" and "tfcommand" directories get created as expected.

I was able to test this in docker using the following command:
```
docker run -d -p 8180:8180 -v greg1.yml:/config/ema.yml --env JAVA_OPTS="-Dapp.commandroot=/opt" --name event-management-agent event-management-agent:ema-test1
```
The "magic" here is setting the environment variable `JAVA_OPTS` to be `-Dapp.commandroot=/opt`. This overrides the use of `user.home`. For the kubernetes deployment, we must add the `JAVA_OPTS` env var via sidekick.

Tested with jar from the command line and with docker container.
Tested with and without the new 'app.commandroot' variable set. With app.commandroot set, it as used as the root directory instead of the user's home directory.
Tested with config push and scan

None

### DIFF
--- a/service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/configuration/TerraformProperties.java
+++ b/service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/configuration/TerraformProperties.java
@@ -7,6 +7,6 @@ import org.springframework.stereotype.Service;
 @Service
 @Data
 public class TerraformProperties {
-    @Value("${COMMAND_PATH:${user.home}${file.separator}tfcommands}")
+    @Value("${app.commandroot:${user.home}}${file.separator}tfcommands")
     private String workingDirectoryRoot;
 }
--- a/service/application/src/main/resources/command-configs.properties
+++ b/service/application/src/main/resources/command-configs.properties
@@ -1,1 +1,1 @@
-COMMAND_PATH=${user.home}${file.separator}commands
+COMMAND_PATH=${app.commandroot:-${user.home}}${file.separator}commands


service/terraform-plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/terraform/configuration/TerraformProperties.java
```diff
@@ -7,6 +7,6 @@
 @Service
 @Data
 public class TerraformProperties {
-    @Value("${COMMAND_PATH:${user.home}${file.separator}tfcommands}")
+    @Value("${app.commandroot:${user.home}}${file.separator}tfcommands")
```
moodiRealist, line 10: Do we need to update README instructions to include this new variable for those running EMA locally as a docker container?

gregmeldrum: This issue seems to be specific to the way we are deploying the container in k8s. I'm not sure if it's a rootless deployment or something, but this issue does not occur when running the docker container using the commands in the README.  If you don't specify the `app.commandroot` Spring variable, it defaults to the previous behaviour of using the `user.home` as the root for the `command` and `tfcommand` directories so a user following instructions from our README will see no change.
If you'd like, I can add a section explaining how the new `app.commandroot` spring variable works and how they can use it to customize where the logs and terraform artifacts are created. Let me know and I can add it.

moodiRealist: I see, ok let's leave things this level of flexibility is not needed for local users. If some asks for it we can update the readme of how to do it.

moodiRealist (approved):

PR #192, opened 2024-07-15
[DATAGO-80317] fix issues blocking standalone mode from executing. 

moodiRealist:
To fix bugs when running the application in standalone mode and certain config values are not available.

Made the existence of a troublesome bean optional.
Also update the default for `event-portal.gateway.messaging.standalone` in conditionalExpressions to set to True.

Updated tests, and manually created docker image, then ran scans from both standalone and connected mode. They both worked.

Connected:
-----------------
Screenshot 2024-07-16 at 3 53 16 PM
Standalone:
-----------------
Screenshot 2024-07-16 at 3 42 43 PM

Nope.


service/application/src/test/java/com/solace/maas/ep/event/management/agent/scanManager/ScanManagerHandleErrorTest.java
mynecker, line 1: @moodiRealist I added a second test to cover standalone mode, too

mynecker (approved):
LGTM and @moodiRealist please cross-check the added test case

PR #193, opened 2024-07-18
DATAGO-80463: micrometer statsd log spam

mynecker:
Fixing log spam of micrometers statsd exporter when `management.statsd.metrics.export.enabled: false` in `env.yml` is not provided.

Added SpringBoot default configuration in `EMAApplication.java` (lowest precedence)

In local development environment
- created a Docker image 
- started EMA without any `management.micrometer.*` configurations defined in `env.yml`
- could not observer log spam anymore

- started EMA with configuration `management.statsd.metrics.export.enabled: true` in `env.yml`
- log spam can be observed - as expected due to missing `statsd` UDP listener

nope

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/EMAApplication.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/EMAApplication.java
@@ -51,6 +51,8 @@ public class EMAApplication {
 
         properties.put("spring.application.name", "event-management-agent");
         properties.put("server.port", "8180");
+        properties.put("management.statsd.metrics.export.enabled", "false");
         return properties;
     }
 }
+


service/application/src/main/java/com/solace/maas/ep/event/management/agent/EMAApplication.java
```diff
@@ -51,6 +51,8 @@ public static Properties getDefault() {
 
         properties.put("spring.application.name", "event-management-agent");
         properties.put("server.port", "8180");
+        properties.put("management.statsd.metrics.export.enabled", "false");
```
moodiRealist, line 54: How did 007 squad use this feature btw considering that ep-core haven't been updated to generate this config? i.e. the config we send in sidekick request probably wouldn't have this value set at all!

moodiRealist: we talked offline.

moodiRealist (approved):

PR #194, opened 2024-07-18
[DATAGO-79884] Upversion spring-boot from 3.2.5 to 3.2.7 to address a tomcat vulnerability

moodiRealist:
To address a vulnerability and unblock EMA from getting released.

Upversioned spring-boot to 3.2.7 which has the newer version of tomcat.

Ran the application in both docker and jar modes.

    ...


mynecker (approved):
LGTM and cross checked in local dev environment

PR #196, opened 2024-08-12
DATAGO-81660: event-management-agent Dockerfile must account for correct user and user home directory setup

rudraneel-chakraborty:

Implement asks of https://sol-jira.atlassian.net/browse/DATAGO-81660

By changing the docker file 

- Created a user called `emauser` (Note: the username can be overiddeen by passing the arfument to the Dockerfile)
- `.terraformrc` file is copied to the `/home/emauser` directory
- switched to using non-root user `emauser` 

Tested configpush and scan for : 
- self managed ema 
- cloud managed ema 

    ...

### DIFF
--- a/service/application/docker/Dockerfile
+++ b/service/application/docker/Dockerfile
@@ -12,7 +12,7 @@ RUN adduser -D $USER && \
     mkdir -p /opt/ema && \
     chmod 777 /opt/ema && \
     mkdir -p /opt/ema/terraform && \
-    chmod 777 /opt/ema/terraform &&\
+    chmod 777 /opt/ema/terraform && \
     chown -R $USER:$USER $HOME && \
     chown -R $USER:$USER /opt/ema/    
 
--- a/service/application/docker/Dockerfile
+++ b/service/application/docker/Dockerfile
@@ -7,13 +7,8 @@ ENV HOME /home/$USER
 
 VOLUME /tmp
 
-# install sudo as root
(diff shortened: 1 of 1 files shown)


service/application/docker/Dockerfile
```diff
@@ -1,8 +1,20 @@
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE}
 
+# default user name is "emauser"
+ARG USER=emauser
```
rudraneel-chakraborty, line 5: Note: you can override the username by `--build-arg USER=someuser`

moodiRealist (approved):

service/application/docker/Dockerfile
```diff
@@ -1,26 +1,39 @@
...
 
 COPY tofu_1.7.0_amd64.apk /opt/ema/terraform
 RUN apk --update add --allow-untrusted /opt/ema/terraform/tofu_1.7.0_amd64.apk
 
-ARG SOLACE_PROVIDER_VERSION=0.9.2-rc.2
-ADD terraform-provider-solacebroker_${SOLACE_PROVIDER_VERSION}_${PLATFORM}.tar.gz /opt/ema/terraform
-
-COPY .terraformrc /root/.terraformrc
+# switch to the created user from root
+USER $USER
```
rudraneel-chakraborty, line 27: This is the earliest we can switch to user as line 24 requires to use `apk`

PR #197, opened 2024-08-13
[DATAGO-78316] Upversion solace-messaging-client to a version that supports partitioned queues

moodiRealist:
EMA must be able to bound to solace queue partitions on E-VMR.

Updated the version of solace-messaging-client to a version that supports partitioned queues

Manually deployed the image to ep-perf-dev and confirmed both ConfigPush and Scan flows work.

The datacenter page deployment used to fail with the existing images after Queues became partitioned, they work now with this image!

Screenshot 2024-08-12 at 3 33 33 PM

Nope.

### DIFF
--- a/service/plugin/pom.xml
+++ b/service/plugin/pom.xml
@@ -107,7 +107,12 @@
         <dependency>
             <groupId>com.solace</groupId>
             <artifactId>solace-messaging-client</artifactId>
-            <version>1.0.0</version>
+            <version>1.4.0</version>
+        </dependency>
+        <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
+            <version>2.6</version>
         </dependency>
         <dependency>
             <groupId>com.solacesystems</groupId>
--- a/service/solace-plugin/pom.xml
+++ b/service/solace-plugin/pom.xml
@@ -10,7 +10,7 @@
     <properties>
         <maven.compiler.source>17</maven.compiler.source>
         <maven.compiler.target>17</maven.compiler.target>
-        <solace-messaging-client.version>1.0.0</solace-messaging-client.version>
+        <solace-messaging-client.version>1.4.0</solace-messaging-client.version>
         <solclientj.version>10.0.0</solclientj.version>
         <spring-boot.version>3.2.7</spring-boot.version>
         <jupiter.version>5.10.2</jupiter.version>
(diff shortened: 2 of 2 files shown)


service/plugin/pom.xml
```diff
@@ -107,7 +107,12 @@
         <dependency>
             <groupId>com.solace</groupId>
             <artifactId>solace-messaging-client</artifactId>
-            <version>1.0.0</version>
+            <version>1.4.0</version>
+        </dependency>
+        <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
+            <version>2.6</version>
```
rudraneel-chakraborty, line 115: Why are we using an older version?? 
https://mvnrepository.com/artifact/org.apache.commons/commons-lang3/3.16.0

service/solace-plugin/pom.xml
```diff
@@ -40,6 +40,11 @@
             <artifactId>solace-messaging-client</artifactId>
             <version>${solace-messaging-client.version}</version>
         </dependency>
+        <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
+            <version>2.6</version>
+        </dependency>
```
rudraneel-chakraborty, line 47: Same comment 
This is the newer version and I think we MUST switch to this : https://mvnrepository.com/artifact/org.apache.commons/commons-lang3/3.16.0

service/plugin/pom.xml
moodiRealist: This is not an older version, but a different library. The one you mentioned is lang3 and the code has been using lang2. After up-versioning the `solace-messaging-client` library from 1.0.0 to 1.4.0 , it stopped bringing in the `commons-lang` dependency for plugins that are using it.

Since the amount of refactoring to completely remove lang2 was substantial and would change the content of this PR from 2 file change to 20+, I decided to not apply all those refactorings on this irrelevant PR.

rudraneel-chakraborty (approved):

service/plugin/pom.xml
rudraneel-chakraborty: I understand , however, please note the older library has CVE vulnerability and we need to fix this

moodiRealist: I don't see any vulnerability reported here https://mvnrepository.com/artifact/commons-lang/commons-lang/2.6 , where did you see them?

An example of one `with` vulnerabilities https://mvnrepository.com/artifact/org.apache.commons/commons-lang3/3.2

PR #199, opened 2024-08-21
DATAGO-78976: Quality | EMA metrics for number of ongoing jobs + health check metrics

CameronRushton:

Add metrics for scan, config push, and heartbeat where it is easy to do, so that POs and devs can see how many requests are going through the system and if the EMA is connected to the EVMR via heartbeat.

Adding counters on events sent and received for CEMA scans and config push. I'll say that again, it's cloud only.
Added a gauge for if the CEMA is connected to the EVMR. Returns a 1 if connected, 0 otherwise.
Added cycle time for config push since it was easy to do so. The request object exists the entire way until it gets converted into a response.
Scan is harder to implement because the initial request is lost in some mapping and async behaviour.

Deploying and running the CEMA for a scan and config push in ep-perf.

None


service/application/src/main/java/com/solace/maas/ep/event/management/agent/service/ScanService.java
```diff
@@ -62,11 +69,17 @@ public class ScanService {
...
-                       RouteService routeService, ProducerTemplate producerTemplate,
-                       IDGenerator idGenerator) {
+                       ScanTypeRepository scanTypeRepository,
+                       ScanStatusRepository scanStatusRepository,
+                       ScanRouteService scanRouteService,
+                       RouteService routeService,
+                       ProducerTemplate producerTemplate,
+                       IDGenerator idGenerator,
+                       MeterRegistry meterRegistry) {
         this.repository = repository;
```
rudraneel-chakraborty, line 83: 👍

rudraneel-chakraborty (approved):

PR #200, opened 2024-08-27
DATAGO-83040: Support Rolling Update

rudraneel-chakraborty:

    ...

    ...

    ...

    ...

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/config/SolaceConfiguration.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/config/SolaceConfiguration.java
@@ -14,6 +14,7 @@ import com.solace.messaging.config.profile.ConfigurationProfile;
 import com.solace.messaging.publisher.DirectMessagePublisher;
 import com.solace.messaging.publisher.OutboundMessageBuilder;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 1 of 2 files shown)


service/application/src/main/java/com/solace/maas/ep/event/management/agent/config/SolaceConfiguration.java
```diff
@@ -114,4 +120,17 @@ public SolacePublisher solacePublisher() {
...
+        try {
+            hostName = InetAddress.getLocalHost().getHostName();
+        } catch (UnknownHostException e) {
+            log.warn("Could not determine host name", e);
+            return StringUtils.EMPTY;
+        }
+
+        String hostNameHash = DigestUtils.sha256Hex(hostName);
+        String agentId = eventPortalProperties.getRuntimeAgentId();
+        return String.format("%s-%s", agentId, hostNameHash);
```
rudraneel-chakraborty, line 134: Hash because 
image

service/application/src/main/java/com/solace/maas/ep/event/management/agent/config/SolaceConfiguration.java
```diff
@@ -62,10 +66,12 @@ public String getTopicPrefix() {
     @ConditionalOnMissingBean(EnableRtoCondition.class)
     @ConditionalOnProperty(name = "event-portal.gateway.messaging.standalone", havingValue = "false")
     public MessagingService messagingService() {
-        String clientName = "runtimeAgent-" + eventPortalProperties.getRuntimeAgentId();
+        String computedClientName = determineClientName();
+        String clientName = StringUtils.isEmpty(computedClientName) ?
+                "runtimeAgent-" + eventPortalProperties.getRuntimeAgentId() : computedClientName;
```
rudraneel-chakraborty, line 71: Falls back on the ep core generated client name if EMA is having trouble determining its own client name based on host name. 
In other words, we are following precedence here
- EMA tries to generate its own client name 
- If it cannot, then it uses the one generated by ep core

PR #202, opened 2024-09-03
DATAGO-83861: ConfigPush Dangling Resource| Scan supporting persistent messaging

rudraneel-chakraborty:

    ...

    ...

    ...

    ...


service/application/src/main/java/com/solace/maas/ep/event/management/agent/subscriber/messageProcessors/ScanCommandMessageProcessor.java
```diff
@@ -68,7 +73,33 @@ public void processMessage(ScanCommandMessage message) {
...
+                        } catch (Exception e) {
+                            log.error("Error while waiting for scan to complete", e);
+                            return false;
+                        }
+                    });
+        } catch (ConditionTimeoutException e) {
+            // Handle the timeout scenario as needed
+            log.error("Scan with id {} did not complete within the expected time", scanId);
+        }
+    }
```
rudraneel-chakraborty, line 99: IMO, we should only do this if CEMA is running on managed mode. It is very easy to do so I will get back to this

moodiRealist:
closing this.

PR #204, opened 2024-09-19
DATAGO-85215: EMA generates its own client name based on hostname and runtime agent id

rudraneel-chakraborty:

To support rolling update, we want each EMA instance to have its own unique `clientName`. Currently, the clientName is determined by only looking at the runtimeAgentId however, this becomes a problem for CEMA update scenario as, for a brief period, we can have more than 1 instance of CEMA with the same id online at the same time. Not to mention, this is also a problem if we decide to use multiple instances (replicaCount >1).

This PR introduces changes so that CEMA generates its own unique client name based on the hash of the host name and appending the runtime agent id. Format is : 

`{runtimeAgentId}-{hash(hostname)}`.

We are taking the hash to ensure that we are not going over the limit 
image

Java

IT

    ...


moodiRealist (approved):
Let's `not merge` this until we have an ep-core PR we can test both of them together in a dev cluster (or ep-perf)

PR #205, opened 2024-09-23
DATAGO-83401: Upgrade spring boot to 3.3.4

bol444:

Upgrade spring boot version to 3.3.4 for EMA

Updated the root pom and the pom of each plugin/module

- Unit tests are passing ✅ 
- Manual testing has been performed as outlined below ✅ 
-- Solace-broker-S_EMA-Mac 
Solace-self-Mac

-- Solace-broker-S_EMA-Windows
Solace-self-Win

-- Kafka-self-Mac 
Kafka-self-Mac

-- Solace-broker-C_EMA
solace-cloud-DD
solace-cloud-audit

none yet


moodiRealist (approved):
We tested on all different brokers and also for Cloud EMA.

PR #207, opened 2024-09-25
Pull a certain image tag when going to anywhere but dev

CameronRushton:

There's a scenario where we want to deploy version X.Y.Z.
We need to run different deploy scripts to go to dev, staging, then prod. This may take several days for testing in between.
In each case, we pull the image tagged with main which is whatever is in the main branch.
For dev, this is good. For staging and production, we only want to deploy the version we said we would, so the version we deploy is also the image we pull.

If we deployed to dev tagged as 1.14, then committed a new change and made the latest release 1.15, the next time we deploy to staging and prod will pull the image tagged with main which is 1.15 and then tagged with 1.14. This is confusing. 

Grab the correct image version instead of main when no deploying to development.

We're testing on the fly.

None

### DIFF
--- a/.github/workflows/deploy-managed-ema-image.yaml
+++ b/.github/workflows/deploy-managed-ema-image.yaml
@@ -3,7 +3,7 @@ on:
   workflow_dispatch:
     inputs:
       releaseVersion:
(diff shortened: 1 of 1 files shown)


moodiRealist (approved):
LGTM

PR #212, opened 2024-11-20
DATAGO-85698: Jira tickets should be created from C-EMA error logs

CameronRushton:

Add the CEMA ID to the logs so that it shows up in our monitoring

Add the EMA ID to MDC

Deploying a copy of the EMA and testing the monitor in dev

None

### DIFF
--- a/service/application/src/main/resources/logback-spring.xml
+++ b/service/application/src/main/resources/logback-spring.xml
@@ -26,7 +26,7 @@
         <if condition='property("USE_JSON").equals("true")'>
             <then>
                 <encoder class="net.logstash.logback.encoder.LogstashEncoder">
-                    <customFields>{"component":"event-management-agent-${agentId}"}</customFields>
+                    <customFields>{"component":"event-management-agent-${agentId}", "ema_id":"${agentId}"}</customFields>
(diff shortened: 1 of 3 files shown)


service/application/src/main/resources/logback-spring.xml
```diff
@@ -26,7 +26,7 @@
         <if condition='property("USE_JSON").equals("true")'>
             <then>
                 <encoder class="net.logstash.logback.encoder.LogstashEncoder">
-                    <customFields>{"component":"event-management-agent-${agentId}"}</customFields>
+                    <customFields>{"component":"event-management-agent-${agentId}", "ema_id":"${agentId}"}</customFields>
```
moodiRealist, line 29: Since the `agentId` is already mentioned in the `component` value, can we use that value instead?

Merging this change would mean, releasing a new EMA after running tests, updating all production C-EMAs. So it would be great if we can use the existing value from `component` instead.

CameronRushton: We absolutely can. That would have saved me a lot of time. I'll edit and re-test the monitor.

PR #215, opened 2025-01-07
DATAGO-90872: Make scan work with Public CEMA

rudraneel-chakraborty:

    ...

    ...

    ...

    ...


service/application/src/main/resources/application.yml
```diff
@@ -4,393 +4,52 @@ springdoc:
     path: /docs/event-management-agent
```
rudraneel-chakraborty, line 4: Ignore this file

PR #217, opened 2025-01-13
DATAGO-87326: update spring-context to address vulnerability

moodiRealist:
update spring-context to address vulnerability

Updated a bunch of dependencies that were pulling in spring-context:6.1.13

- S-EMA
  - Solace
    - Cloud 
S-EMA solace configPush

    - Appliance 
appliance changes work

  - kafka 
kafka scan works

- C-EMA 
C-EMA functional tests all passed
C-EMA scan works

- Windows 
Screenshot 2025-01-14 221322

    ...

### DIFF
--- a/service/kafka-plugin/pom.xml
+++ b/service/kafka-plugin/pom.xml
@@ -15,7 +15,7 @@
         <spring-boot.version>3.3.7</spring-boot.version>
         <commons-collections4.version>4.4</commons-collections4.version>
         <jupiter.version>5.10.2</jupiter.version>
-        <camel.version>4.9.0</camel.version>
+        <camel.version>4.8.3</camel.version>
         <aws-msk-iam-auth.version>2.0.3</aws-msk-iam-auth.version>
         <snakeyaml.version>2.0</snakeyaml.version>
         <jackson-dataformat-cbor.version>2.13.4</jackson-dataformat-cbor.version>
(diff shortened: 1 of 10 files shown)


service/kafka-plugin/pom.xml
```diff
@@ -10,12 +10,12 @@
...
         <maven.compiler.target>17</maven.compiler.target>
-        <spring-kafka.version>3.2.4</spring-kafka.version>
+        <spring-kafka.version>3.2.6</spring-kafka.version>
         <kafka-clients.version>3.5.0</kafka-clients.version>
-        <spring-boot.version>3.3.4</spring-boot.version>
+        <spring-boot.version>3.3.7</spring-boot.version>
         <commons-collections4.version>4.4</commons-collections4.version>
         <jupiter.version>5.10.2</jupiter.version>
-        <camel.version>4.5.0</camel.version>
+        <camel.version>4.8.3</camel.version>
```
moodiRealist, line 18: I am aware that there is also camel dependency `4.9.0` , however that version does a minor upversion on a lot of its own `springboot` dependencies (e.g. instead of spring-context 6.1.16, they use 6.2.0). 

When using 4.8.3, we don't need to worry about a mix of dependencies from versions of springBoot 3.3.7 and Camel 4.8.3 since these two are aligned.

I'm not saying it is impossible to update to 4.9.0, but as you can see this PR is updating a lot of versions already. So I don't want to increase the chance of breaking our features.

CameronRushton (approved):

CameronRushton:
One extra comment since I've already approved the first PR this one is being merged into is that I usually update all the plugin versions as well. It doesn't really matter to do that though.

PR #226, opened 2025-01-22
DATAGO-89835: Public DC C-EMA config push

CameronRushton:

Config push must be able to be performed by several different organizations sharing an EMA.

Use the orgId from the mop message instead of getting it from the app properties to support multi tenancy better.

Manually by running the EMA locally and verifying in the debugger that the org ID was set as expected.

None

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/CommandManager.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/CommandManager.java
@@ -110,7 +110,7 @@ public class CommandManager {
     @SuppressWarnings("PMD")
     private void configPush(CommandRequest request) {
         List<Path> executionLogFilesToClean = new ArrayList<>();
-        boolean attacheErrorToTerraformCommand = false;
+        boolean attachErrorToTerraformCommand = false;
         try {
             // if the serviceId is not found, messagingServiceDelegateService will most likely throw an exception (which is not guaranteed
             // based on the interface definition) and we need to catch it here.
@@ -129,11 +129,11 @@ public class CommandManager {
                 boolean exitEarlyOnFailedCommand = bundle.getExitOnFailure();
                 // For now everything is run serially
                 for (Command command : bundle.getCommands()) {
-                    attacheErrorToTerraformCommand = false;
+                    attachErrorToTerraformCommand = false;
                     if (command.getCommandType() == semp) {
                         executeSempCommand(command, solaceClient);
                     } else if (command.getCommandType() == terraform) {
-                        attacheErrorToTerraformCommand = true;
+                        attachErrorToTerraformCommand = true;
                         Path executionLog = executeTerraformCommand(request, command, envVars);
                         if (executionLog != null) {
                             if (commandLogStreamingProcessorOpt.isPresent()) {
@@ -152,7 +152,7 @@ public class CommandManager {
             }
         } catch (Exception e) {
             log.error("ConfigPush command not executed successfully", e);
-            attachErrorLogToCommand(attacheErrorToTerraformCommand, e, request);
+            attachErrorLogToCommand(attachErrorToTerraformCommand, e, request);
         } finally {
             try {
                 finalizeAndSendResponse(request);
@@ -211,6 +211,7 @@ public class CommandManager {
                 "runtimeAgentId", eventPortalProperties.getRuntimeAgentId(),
                 COMMAND_CORRELATION_ID, requestBO.getCommandCorrelationId()
         );
+        log.info("TEST LOG: orgId of requestBO: {}", requestBO.getOrgId());
         CommandMessage response = new CommandMessage(requestBO.getServiceId(),
                 requestBO.getCommandCorrelationId(),
                 requestBO.getContext(),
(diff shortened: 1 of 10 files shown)


service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/CommandManager.java
```diff
@@ -240,7 +240,7 @@ private Path executeTerraformCommand(CommandRequest request, Command command, Ma
             setCommandError(command, e);
 
         }
-        return executionLog;
+        return null;
```
CameronRushton, line 243: This was always null. IntelliJ recommended this.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/CommandManager.java
```diff
@@ -204,19 +205,19 @@ private void attachErrorLogToCommand(boolean isTerraformRelatedError, Exception
         }
     }
 
-    private void finalizeAndSendResponse(CommandRequest request) {
-        request.determineStatus();
+    private void finalizeAndSendResponse(CommandRequest requestBO) {
+        requestBO.determineStatus();
```
CameronRushton, line 209: Rename from request -> requestBO because it's hard to tell what was part of the messaging layer and what is internal.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/CommandManager.java
```diff
@@ -204,19 +205,19 @@ private void attachErrorLogToCommand(boolean isTerraformRelatedError, Exception
...
         Map<String, String> topicVars = Map.of(
-                "orgId", eventPortalProperties.getOrganizationId(),
+                "orgId", requestBO.getOrgId(),
```
CameronRushton, line 211: No longer grabbing the org ID from the application properties. It now comes from the mop message.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/config/SolaceConfiguration.java
```diff
@@ -48,14 +49,19 @@ public SolaceConfiguration(Properties vmrConfig, ArrayList<String> sessionConfig
...
         }
-        return topicPrefix;
+        return String.format(TOPIC_PREFIX_FORMAT,
+                orgId,
+                eventPortalProperties.getRuntimeAgentId());
+    }
+
+    public String getTopicPrefix() {
+        return getTopicPrefix(null);
     }
```
CameronRushton, line 65: There's several places where we call this to do things like set the config push command subscription and set the topic to send responses to. It's used for scan, config push, and heartbeat. To make this PR low touch, I've only altered anything to do with config push. That's the only case that sends in an orgId right now.
Maybe scan would want to use this as well? cc @rudraneel-chakraborty

service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/CommandLogStreamingProcessor.java
```diff
@@ -111,10 +112,11 @@ private List<String> readAllExecutionLogs(Path path) {
...
-                                 String messagingServiceId) {
+                                 String messagingServiceId,
+                                 String organizationId) {
         try {
             Map<String, String> topicDetails = new HashMap<>();
-            topicDetails.put("orgId", eventPortalProperties.getOrganizationId());
+            topicDetails.put("orgId", organizationId);
             topicDetails.put("runtimeAgentId", eventPortalProperties.getRuntimeAgentId());
             topicDetails.put("messagingServiceId", messagingServiceId);
             topicDetails.put(COMMAND_CORRELATION_ID, commandCorrelationId);
```
CameronRushton, line 122: This should be some kind of object imo but that can be for later.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/publisher/CommandPublisher.java
```diff
@@ -29,12 +31,11 @@ public CommandPublisher(SolacePublisher solacePublisher, SolaceConfiguration sol
...
 
     public void sendCommandResponse(MOPMessage message, Map<String, String> topicDetails) {
-
         String topicString =
                 String.format("%scommandResponse/v1/%s",
-                        solaceConfiguration.getTopicPrefix(),
+                        solaceConfiguration.getTopicPrefix(topicDetails.get("orgId")),
                         topicDetails.get(Command.COMMAND_CORRELATION_ID));
-
+        log.info("Sending config push response to topic: {}", topicString);
```
CameronRushton, line 38: What do we think about logging the topic? It helped with debugging but I wonder if it's too much. I'd like to see a lot of these be debug logs tbh.

rudraneel-chakraborty (changes requested):
Discussed comments with cameron

service/application/src/main/java/com/solace/maas/ep/event/management/agent/publisher/CommandPublisher.java
rudraneel-chakraborty: I don't think we should log the topics, specially in info level

rudraneel-chakraborty (approved):

service/application/src/main/java/com/solace/maas/ep/event/management/agent/publisher/CommandPublisher.java
CameronRushton: I've removed this one and kept the subscription one but lowered that one to debug.

PR #227, opened 2025-01-25
add a way to build the image locally and push it to the dev image rep…

CameronRushton:
…o instead of pulling main from ECR

    ...

    ...

    ...

    ...

### DIFF
--- a/.github/workflows/deploy-managed-ema-image.yaml
+++ b/.github/workflows/deploy-managed-ema-image.yaml
@@ -14,6 +14,10 @@ on:
           - development
           - staging
           - production
+      buildNewDevImage:
+        description: "true/false. If deploying to development, set to 'true' to build a new image, otherwise, we'll pull the 'main' image."
+        required: true
+        default: "A.B.C"
 jobs:
   deploy:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-managed-ema-image.yaml
```diff
@@ -29,37 +28,19 @@ jobs:
...
-      - name: ECR (Dev) - Pull Main Image
-        if: ${{ github.event.inputs.deployEnvironment == 'development' }}
+      - name: Set Build Params
         run: |
-          ECR_DEV_IMAGE="${{ steps.login-ecr.outputs.registry }}/${{ github.event.repository.name }}:main"
-          docker pull $ECR_DEV_IMAGE
-          echo "ECR_DEV_IMAGE=$ECR_DEV_IMAGE" >> $GITHUB_ENV
-      - name: ECR (Dev) - Pull Prod Ready Image Tag
-        if: ${{ github.event.inputs.deployEnvironment != 'development' }}
+          export SKIP_FLAGS_NON_UNIT_TESTS="-Dcheckstyle.skip -Dpmd.skip -Dcpd.skip -Dfindbugs.skip -Dspotbugs.skip"
```
moodiRealist, line 33: Are we adding these to prod since it would make things faster and not needed? or some other reason?

moodiRealist (approved):
LGTM with a small comment to understand a change.

.github/workflows/deploy-managed-ema-image.yaml
CameronRushton: That shouldn't have been in there. Thanks! I removed it.

rudraneel-chakraborty (approved):
Do we need to change Readme or any other documentation?

CameronRushton:
Actually, there may be some small changes to our internal documentation. I'll ensure that's updated.

PR #228, opened 2025-01-27
DATAGO-89836: Support Scan for Public CEMA

rudraneel-chakraborty:

    ...

    ...

    ...

    ...

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/cli/CommandLineScan.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/cli/CommandLineScan.java
@@ -1,5 +1,6 @@
 package com.solace.maas.ep.event.management.agent.cli;
 
+import com.solace.maas.ep.event.management.agent.config.eventPortal.EventPortalProperties;
 import com.solace.maas.ep.event.management.agent.repository.model.mesagingservice.MessagingServiceEntity;
 import com.solace.maas.ep.event.management.agent.scanManager.ScanManager;
 import com.solace.maas.ep.event.management.agent.scanManager.model.ScanRequestBO;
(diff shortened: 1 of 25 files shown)


service/application/src/main/java/com/solace/maas/ep/event/management/agent/cli/CommandLineScan.java
```diff
@@ -45,6 +48,7 @@ public void runScan(String messagingServiceId, String filePathAndName) throws In
                 .messagingServiceId(messagingServiceId)
                 .scanId(idGenerator.generateRandomUniqueId())
                 .destinations(List.of("FILE_WRITER"))
+                .orgId(eventPortalProperties.getOrganizationId())
```
rudraneel-chakraborty, line 51: for command line scan, pass in the org_id from application.yaml

moodiRealist: ok

service/application/src/main/java/com/solace/maas/ep/event/management/agent/service/ScanService.java
```diff
@@ -287,12 +242,21 @@ public Optional<ScanEntity> findById(String scanId) {
...
+    public void sendScanStatus(String orgId,
+                               String groupId,
+                               String scanId,
+                               String traceId,
+                               String actorId,
+                               String messagingServiceId,
+                               String scanTypes,
                                ScanStatus status) {
+
+        Validate.notBlank(orgId, "Organization ID cannot be null or empty");
```
moodiRealist, line 254: I see that this error message have been used in many places. Can we add a CONSTANT for it plz?

PR #236, opened 2025-02-14
Crushton/log is public metric

CameronRushton:

    ...

    ...

    ...

    ...

### DIFF
--- a/.github/workflows/deploy-managed-ema-image.yaml
+++ b/.github/workflows/deploy-managed-ema-image.yaml
@@ -14,10 +14,6 @@ on:
           - development
           - staging
           - production
-      buildNewDevImage:
-        description: "true/false. If deploying to development, set to 'true' to build a new image, otherwise, we'll pull the 'main' image."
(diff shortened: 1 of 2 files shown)


moodiRealist (approved):

moodiRealist (approved):
Approved, but please let @rudraneel-chakraborty have a look as well.

rudraneel-chakraborty (approved):

CameronRushton:
This appears to work correctly for self managed EMAs. The picked path for logback is with `is_public:false` and there's no issues. Though, if the user edited their default application properties to change the org to `*` or changed their org variable to a `*`, the wrong path would then be chosen, marking as `is_public:true`, but there's no reason for them to do this and the field is only used in EMA logs.
The field means nothing to self managed users anyway, though it would be ideal to not have it at all or always be false in this case. There isn't currently a sanctioned way of differentiating EMAs between self managed and in a private DC on the EMA's side and it's not worth trying to do.

PR #238, opened 2025-03-10
DATAGO-96611: Bug Fix, read orgId from application.yaml in standalone mode

rudraneel-chakraborty:

To fix a bug described here: https://sol-jira.atlassian.net/browse/DATAGO-96611

Java

IT

    ...

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/ScanStatusOverAllProcessor.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/ScanStatusOverAllProcessor.java
@@ -7,6 +7,7 @@ import com.solace.maas.ep.event.management.agent.plugin.constants.ScanStatus;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.camel.Exchange;
 import org.apache.camel.Processor;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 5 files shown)


service/application/src/main/java/com/solace/maas/ep/event/management/agent/processor/ScanStatusOverAllProcessor.java
```diff
@@ -45,10 +46,13 @@ public void process(Exchange exchange) throws Exception {
 
         String scanType = (String) properties.get(RouteConstants.SCAN_TYPE);
         String orgId = (String) properties.get(RouteConstants.ORG_ID);
+        if (StringUtils.isEmpty(orgId) && !StringUtils.equals(eventPortalProperties.getOrganizationId(), "*")) {
+            orgId = eventPortalProperties.getOrganizationId();
+        }
```
rudraneel-chakraborty, line 51: Basically, read orgId from application.yml iff
- the context does not have orgId (only happens in manual import mode)
- AND the EMA is not shared (safety)

moodiRealist: Can you please move this condition to a helper method and use that method on other occurrences in this PR?

moodiRealist (approved):

PR #239, opened 2025-03-14
DATAGO-91781: Disable certificate validation for TLS connections between C-EMA and event brokers in private/dedicated regions

CameronRushton:

Skip verifying TLS certs if so desired.

The EMA takes in application properties that will have the skipping TLS boolean in it. That is read and passed to any plugin via ConnectionDetailsEvent. In this case, only the solace plugin makes use of it.

The main application part just uses the application properties.

Added a test in the solace plugin to ensure that a client is created when the new boolean is true or false.
Deployed to a development environment and did a config push and scan against a broker that requires TLS verification to be disabled to be successful.

We can deploy this independently, but best to release this after solace cloud supports it.

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/semp/SempApiProviderImpl.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/semp/SempApiProviderImpl.java
@@ -5,6 +5,7 @@ import com.solace.client.sempv2.api.AclProfileApi;
 import com.solace.client.sempv2.api.AuthorizationGroupApi;
 import com.solace.client.sempv2.api.ClientUsernameApi;
 import com.solace.client.sempv2.api.QueueApi;
+import com.solace.maas.ep.event.management.agent.config.eventPortal.EventPortalProperties;
 import com.solace.maas.ep.event.management.agent.plugin.solace.processor.semp.SempClient;
 import com.solace.maas.ep.event.management.agent.plugin.solace.processor.semp.SolaceHttpSemp;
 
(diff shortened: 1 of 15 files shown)


CameronRushton:
I don't think sonar is able to use the tests in the solace plugin. That section is 'tested' to its standards and shows 0% coverage.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/semp/SempApiProviderImpl.java
```diff
@@ -56,10 +62,14 @@ public QueueApi getQueueApi() {
...
-        apiClient.setUsername(sempClient.getUsername());
-        apiClient.setPassword(sempClient.getPassword());
-        return apiClient;
+        ApiClient client = new ApiClient();
+        client.setBasePath(sempClient.getConnectionUrl() + "/SEMP/v2/config");
+        client.setUsername(sempClient.getUsername());
+        client.setPassword(sempClient.getPassword());
+        boolean verifyTls = eventPortalProperties == null || !eventPortalProperties.getSkipTlsVerify();
+        log.info("SetVerifyingSsl? {} (application properties skipTlsVerify: {})", verifyTls,
+                eventPortalProperties == null ? "false" : eventPortalProperties.getSkipTlsVerify());
```
rudraneel-chakraborty, line 71: I was wondering is this should be a debug log?

service/plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/config/eventPortal/EventPortalPluginProperties.java
```diff
@@ -17,4 +17,5 @@ public class EventPortalPluginProperties {
     private String topicPrefix;
 
     private GatewayProperties gateway;
+    private Boolean skipTlsVerify = false;
```
rudraneel-chakraborty, line 20: is this field getting used anywhere?

CameronRushton: Right. This was the old way I had it but since removed it in favour of the connectionDetailsEvent.

PR #249, opened 2025-04-15
DATAGO-99254: Fix SEMP delete when TLS verification is disabled

CameronRushton:

Fix an issue where SEMP deletes were not working on brokers that have a certificate because we are not disabling verification properly.

Set the properties before setting up the client.

Deploying and running the EMA in a dev environment. Doing a config push, scan, and config delete.

None

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/semp/SempApiProviderImpl.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/semp/SempApiProviderImpl.java
@@ -16,7 +16,6 @@ import lombok.extern.slf4j.Slf4j;
 public class SempApiProviderImpl implements SempApiProvider {
 
     private final ApiClient apiClient;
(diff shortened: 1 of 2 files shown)


service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/semp/SempApiProviderImpl.java
```diff
@@ -25,8 +24,7 @@ public class SempApiProviderImpl implements SempApiProvider {
 
     public SempApiProviderImpl(SolaceHttpSemp solaceClient,
                                EventPortalProperties eventPortalProperties) {
-        this.apiClient = setupApiClient(solaceClient);
-        this.eventPortalProperties = eventPortalProperties;
```
CameronRushton, line 29: I set the properties after by mistake :(

moodiRealist (approved):

PR #262, opened 2025-06-05
DATAGO-102788: Accept broker URLs with extended SempUri path

moodiRealist:
To unblock customers with extended sempUri path for their broker URL.

Updated the SolaceHttpSemp logic

Added UntiTest coverage.

Also, ran the EMA as S-EMA docker container and observed the following:

- Things work as expected for customers with simple broker URLs: 

regular URL
regular URl - docker logs

- If we provide extended URL, all of it is taken into account and we don't ignore the provided path:

new URL
new URl - docker logs

    ...


bol444 (approved):
LGTM, thank you, Moodi

PR #263, opened 2025-06-10
DATAGO-103686: Scan for federated account POC 

rudraneel-chakraborty:

    ...

    ...

    ...

    ...


service/application/src/main/java/com/solace/maas/ep/common/messages/ScanDataMessage.java
```diff
@@ -8,24 +8,22 @@
...
         withMessageType(MOPMessageType.generic)
                 .withProtocol(MOPProtocol.scanData)
                 .withVersion("1")
                 .withUhFlag(MOPUHFlag.ignore);
 
         this.orgId = orgId;
+        this.originOrgId = originOrgId;
         this.scanId = scanId;
         this.scanType = scanType;
         this.data = data;
```
rudraneel-chakraborty, line 29: Refer to the comment https://github.com/SolaceDev/maas-ep-core/pull/3802/files#r2140995936
we should be using Lombok annotations and not rely on ordering in the constructor as this is error prone and can easily introduce bugs

PR #301, opened 2026-05-26
DATAGO-137993: add SEMP delete support for ACL client connect exception

bol444:

Adds EMA-side support for the SEMP DELETE of an ACL client connect exception (IP whitelist entry). This is the EMA half of a two-repo coordinated change; the matching ep-core change is tracked under DATAGO-134695 (https://sol-jira.atlassian.net/browse/DATAGO-134695).

Mirror the existing publish/subscribe topic exception delete pattern exactly. Four touches:

  - add solaceAclClientConnectException enum value.
  - SempAclClientConnectExceptionDeletionRequest.java (new) — POJO with msgVpn, aclProfileName, clientConnectExceptionAddress. 
    - No topicSyntax field — unlike topic exceptions, the SEMP API for client connect exception takes 3 path params, not 4.
  - SempDeleteCommandManager.java — new switch case + executeDeleteAclClientConnectException(...) handler. Same shape as executeDeleteAclPublishTopicException

  - ITs
  - End-to-end on ep-perf will be exercised once the matching ep-core PR lands behind its feature flag: push [1.2.3.4/32, 5.6.7.8/32] → push [5.6.7.8/32] → SEMP GET confirms only 5.6.7.8/32 remains.

- EMA changes must be in prod before ep-core starts emitting solaceAclClientConnectException commands. 
- An older EMA receiving this command type will throw IllegalArgumentException: Unsupported SEMP delete entity
  type and fail the entire config push.

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/SempDeleteCommandManager.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/SempDeleteCommandManager.java
@@ -112,7 +112,7 @@ public class SempDeleteCommandManager extends AbstractSempCommandManager {
                 objectMapper.writeValueAsString(command.getParameters().get(SempCommandConstants.SEMP_COMMAND_DATA)),
                 SempAclProfileDeletionRequest.class);
         Validate.notEmpty(request.getMsgVpn(), MSG_VPN_EMPTY_ERROR_MSG);
-        Validate.notEmpty(request.getAclProfileName(), "ACL profile name must not be empty");
+        Validate.notEmpty(request.getAclProfileName(), ACL_PROFILE_NAME_EMPTY_ERROR_MSG);
 
         log.info("SEMP delete: Deleting ACL profile");
         try {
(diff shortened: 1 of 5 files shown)


service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/SempDeleteCommandManager.java
```diff
@@ -93,6 +94,9 @@ protected void executeSempCommand(Command command, SempApiProvider sempApiProvid
             case solaceAclPublishTopicException:
                 executeDeleteAclPublishTopicException(command, sempApiProvider);
                 break;
+            case solaceAclClientConnectException:
+                executeDeleteAclClientConnectException(command, sempApiProvider);
+                break;
```
bol444, line 99: added a new case to handle `solaceAclClientConnectException`, following the existing pattern of `solaceAclPublishTopicException` & `solaceAclSubscribeTopicException`

service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/SempDeleteCommandManager.java
```diff
@@ -108,7 +112,7 @@ private void executeDeleteAclProfile(Command command, SempApiProvider sempApiPro
                 objectMapper.writeValueAsString(command.getParameters().get(SempCommandConstants.SEMP_COMMAND_DATA)),
                 SempAclProfileDeletionRequest.class);
         Validate.notEmpty(request.getMsgVpn(), MSG_VPN_EMPTY_ERROR_MSG);
-        Validate.notEmpty(request.getAclProfileName(), "ACL profile name must not be empty");
+        Validate.notEmpty(request.getAclProfileName(), ACL_PROFILE_NAME_EMPTY_ERROR_MSG);
```
bol444, line 115: abstracted this string into a constant

charberg (approved):

PR #305, opened 2026-06-05
DATAGO-138201: suppress broker INVALID_PARAMETER on RDP queue-binding request-header DELETE

bol444:

EMA's config push deletes RDP queue-binding request headers via SEMP.  When a user configures a header name the broker doesn't accept (e.g. `Authorization`, which is reserved), the first push fails on the CREATE. That part is expected.
The problem is the recovery push (this is the reported bug). When the user removes the bad header and pushes again, ep-core asks EMA to delete the stale header from the broker. The broker rejects this DELETE with `400 + status: "INVALID_PARAMETER"`.  EMA only knows how to swallow `400 + NOT_FOUND`, so it rethrows and the push fails.

Note: based on manual testing, a third push then succeeds on its own, but the customer still has to live through one failed recovery push. This fix removes that.

Bug link: https://sol-jira.atlassian.net/browse/DATAGO-138201

Changed `SempDeleteCommandManager.executeRdpQueueBindingRequestHeader` (regular + protected variants). After the DELETE call, two broker responses are now suppressed instead of one:

  - `400 + NOT_FOUND` — already handled by the shared `handleSempOperationException`. Unchanged.
  - `400 + INVALID_PARAMETER` — new. Logged at INFO with entity type only. Returns success.

Anything else still rethrows.

Manual test on ep-perf: push 1 with bad header fails as expected. Push 2 with the header removed used to fail on the SEMP DELETE — now it succeeds.

**Other DELETE surfaces** in `SempDeleteCommandManager` (ACL profile, etc.) may have the same class of bug (flagged by Claude, but I didn't test yet), will investigate separately

### DIFF
--- a/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/SempDeleteCommandManager.java
+++ b/service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/SempDeleteCommandManager.java
@@ -325,45 +325,23 @@ public class SempDeleteCommandManager extends AbstractSempCommandManager {
         Validate.notEmpty(request.getRdpName(), "RDP name must not be empty");
         Validate.notEmpty(request.getQueueBindingName(), "Queue binding name must not be empty");
 
-        // Pre-flight GET: skip the DELETE if the header is not on the broker.
-        // The broker can return INVALID_PARAMETER (not NOT_FOUND) when deleting a
-        // header that was recorded but never accepted on an earlier CREATE.
+        String entityType = request.isProtected() ? "Protected Request Header" : "Request Header";
         try {
             if (request.isProtected()) {
-                restDeliveryPointApi.getMsgVpnRestDeliveryPointQueueBindingProtectedRequestHeader(
-                        request.getMsgVpn(), request.getRdpName(),
-                        request.getQueueBindingName(), request.getHeaderName(),
-                        null, null);
-            } else {
-                restDeliveryPointApi.getMsgVpnRestDeliveryPointQueueBindingRequestHeader(
-                        request.getMsgVpn(), request.getRdpName(),
-                        request.getQueueBindingName(), request.getHeaderName(),
-                        null, null);
-            }
-        } catch (ApiException e) {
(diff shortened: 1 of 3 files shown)


service/application/src/main/java/com/solace/maas/ep/event/management/agent/command/SempDeleteCommandManager.java
```diff
@@ -324,23 +324,24 @@ private void executeRdpQueueBindingRequestHeader(Command command, SempApiProvide
...
                 restDeliveryPointApi.deleteMsgVpnRestDeliveryPointQueueBindingRequestHeader(request.getMsgVpn(), request.getRdpName(),
                         request.getQueueBindingName(), request.getHeaderName());
-            } catch (ApiException e) {
-                handleSempOperationException(e, "Request Header", supportedSempCommand());
             }
+        } catch (ApiException e) {
+            if (e.getCode() == 400 && e.getResponseBody() != null && e.getResponseBody().contains("INVALID_PARAMETER")) {
+                log.info("SEMP {}: broker rejected {} identifier as invalid; treating as not present", supportedSempCommand(), entityType);
+                return;
+            }
```
CameronRushton, line 343: Shouldn't this be moved into handleSempOperationException?

bol444: that's a good question, @CameronRushton . `handleSempOperationException` is called from every DELETE in the file (executeDeleteAclProfile, executeDeleteSolaceQueue, etc.), and having this update inside `handleSempOperationException` means we are suppressing broker INVALID_PARAMETER for all entities, including ACL profile and queue etc., which is much larger testing effort, so I limited the fixed to only on rdp queue binding request header

CameronRushton: Got it. Thanks

CameronRushton (approved):

PR #312, opened 2026-06-24
DATAGO-141720: Automate S-EMA publish to DockerHub

bol444:
## Summary

Adds `.github/workflows/publish-to-dockerhub.yaml` to automate **Phase 6** of the EMA prod release — the S-EMA image push to public DockerHub `solace/event-management-agent`. This is currently the only fully-manual hop in the release pipeline (local command line based `docker pull → tag → push`); every other hop is already gated by GitHub Actions + the `prod` environment.

Jira: https://sol-jira.atlassian.net/browse/DATAGO-141720

## Today's manual flow

1. Operator logs into AWS ECR + DockerHub on their laptop.
2. `docker pull` ECR `:<version>` → `docker tag` for `solace/...:<version>` + `:latest` → `docker push` both.
3. Verify in browser.

## What are the changes

Mirrors the existing `deploy-managed-ema-image.yaml` pattern:

- `workflow_dispatch` with two inputs: `releaseVersion` (required) and `pushLatest` (boolean, default false).
- `environment: prod` for both the second-reviewer approval gate and the secret scoping.
- `aws-actions/configure-aws-credentials`, `aws-actions/amazon-ecr-login`, `docker/login-action` — all SHA-pinned to the same versions referenced by `deploy-managed-ema-image.yaml`.
- `permissions: contents: read` (minimum).

Additional guards beyond the existing pattern:

- **Strict semver validation** rejects `v1.9.9`, `1.9.9.0`, `1.9`, etc. before any AWS / DockerHub calls.
- **`:latest` pre-release guard** — any `releaseVersion` carrying a suffix (`A.B.C-...`) is refused when `pushLatest=true`. Stricter than an allowlist of suffixes, deliberately, so the guard doesn't drift as new release flavours appear.
- **ECR-existence pre-flight** — `aws ecr describe-images` before `docker pull` so a missing tag surfaces as a human-readable error rather than the opaque `manifest unknown`.
- **Per-version `concurrency:` block** with `cancel-in-progress: false` — prevents two operators racing a push of the same version. The deliberate \"don't cancel mid-flight\" choice avoids corrupt layer uploads on DockerHub.
- **Idempotency observability** — warns (does not block) when re-pushing a `:<version>` tag that already exists. Re-push is safe (ECR release tags are immutable, same bits get written); the warning makes the situation legible in the audit summary.
- **`if: always()` audit summary** with per-step outcomes (`success` / `failure` / `skipped` / `not-reached`). Reports what actually happened rather than what we intended to do.

### Why a separate workflow, not a third branch of `deploy-managed-ema-image.yaml`

Considered. Rejected because:

- DockerHub pushes two tags (`:version` + `:latest`); GCR pushes one. Conditional logic would muddy the existing file.
- The `:latest` pre-release guard has no analog on the GCR side.
- Audit trail is clearer when a DockerHub publish is its own workflow run, not a sub-branch of a multi-target invocation.

## Secret scoping (suggested by Claude, I think it is well reasoned)

`DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` should be added as **`prod`-environment secrets**, not repo-level secrets.

Rationale: repo-level secrets are readable by **any** workflow in the repo on any branch. That includes workflows whose third-party-action list is not fully SHA-pinned. For example, `release.yaml:137` references `Bullrich/generate-release-changelog@master` — pinned to a moving Git ref. If that action is compromised, the malicious code runs inside `release.yaml`'s `publish` job (which already has `environment: prod`) and can read any repo-level secret. A `prod`-environment secret is only injected into a job that declares `environment: prod`, and only **after** the prod approval gate passes — bounding the blast radius to approved runs.

The new workflow declares `environment: prod`, so it will read the secrets correctly once they're added.

### Service-account DockerHub user — does one exist, and is the pattern endorsed?

The plan calls for the PAT to live on a **dedicated bot account**, not on a personal DockerHub user. Reasons:

- Decouples the credential from any individual person's lifecycle
- Scopes the PAT to **only** what `eventportalrelease` needs (least privilege).

Open sub-questions: Does Solace already have a DockerHub bot user we should reuse, or do we create a new one?

The eventual onboarding ask (add bot user to `solace/eventportalrelease` team, generate `Read & Write`-scoped PAT, register secrets in the `prod` environment) is held until the team agrees on the approach.

### DIFF
--- a/.github/workflows/publish-to-dockerhub.yaml
+++ b/.github/workflows/publish-to-dockerhub.yaml
@@ -1,0 +1,162 @@
+name: Publish EMA to DockerHub
+on:
+  workflow_dispatch:
+    inputs:
+      releaseVersion:
+        description: "Released version to publish (must already exist as ECR :<version>)."
+        required: true
+        default: "A.B.C"
+      pushLatest:
+        description: "Also overwrite the :latest tag. Only enable for stable releases."
+        type: boolean
+        default: false
+
+# Prevent two operators (or one double-click) from racing to publish the same
+# version. cancel-in-progress: false is deliberate — cancelling a docker push
+# mid-flight can leave a corrupt layer upload on DockerHub. The second run
+# queues, waits for the first to finish, then fails when the tag already
+# exists.
+concurrency:
+  group: publish-dockerhub-${{ inputs.releaseVersion }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    environment: prod          # manual approval gate, identical to deploy-managed-ema-image.yaml
+
+    steps:
+      # Two-step version validation. Step A always runs and catches typos
+      # early (before any AWS or DockerHub calls). Step B enforces the
+      # "stable release = no suffix" invariant when pushLatest is set.
+      - name: Validate releaseVersion shape
+        run: |
+          VERSION="${{ inputs.releaseVersion }}"
+          # Step A: strict semver. Allow A.B.C or A.B.C-<alphanumeric suffix>.
+          if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[A-Za-z0-9]+)?$ ]]; then
+            echo "::error::releaseVersion '$VERSION' is not a valid semver (expected A.B.C or A.B.C-<suffix>)."
+            exit 1
+          fi
+          # Step B: any suffix disqualifies the version from updating :latest.
+          # This is intentionally stricter than an allowlist of suffixes — it
+          # avoids the maintenance burden of keeping the list current as new
+          # release flavours (e.g. -hotfix, -cr, -customer1) are introduced.
+          if [[ "${{ inputs.pushLatest }}" == "true" ]] && [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+- ]]; then
+            echo "::error::Refusing to push :latest for pre-release version '$VERSION'. Stable releases have no suffix."
+            exit 1
+          fi
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@5fd3084fc36e372ff1fff382a39b10d03659f355  # v2.2.0
+        with:
+          aws-access-key-id: ${{ secrets.EMA_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.EMA_AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.EMA_AWS_DEFAULT_REGION }}
+
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@2f9f10ea3fa2eed41ac443fee8bfbd059af2d0a4  # v1.6.0
+
+      # Pre-flight: surface "ECR tag missing" as a human-readable error
+      # before we attempt the docker pull (whose 'manifest unknown' failure
+      # is opaque). The most likely cause of a missing tag is that the
+      # Release workflow's publish job hasn't run yet for this version.
+      - name: Verify ECR tag exists
+        env:
+          REGION: ${{ secrets.EMA_AWS_DEFAULT_REGION }}
+          REPO: ${{ github.event.repository.name }}
+          VERSION: ${{ inputs.releaseVersion }}
+        run: |
+          if ! aws ecr describe-images \
+               --repository-name "$REPO" \
+               --image-ids imageTag="$VERSION" \
+               --region "$REGION" \
+               > /dev/null 2>&1; then
+            echo "::error::ECR tag '$REPO:$VERSION' not found in region $REGION."
+            echo "::error::The Release workflow's publish job must run successfully first to create this tag."
+            echo "::error::See: code_bases/ema/deployment/concepts/image-flow.md"
+            exit 1
+          fi
+          echo "::notice::Confirmed ECR tag '$REPO:$VERSION' exists."
+
+      - name: Pull image from ECR
+        run: |
+          ECR_IMAGE="${{ steps.login-ecr.outputs.registry }}/${{ github.event.repository.name }}:${{ inputs.releaseVersion }}"
+          docker pull "$ECR_IMAGE"
+          echo "ECR_IMAGE=$ECR_IMAGE" >> $GITHUB_ENV
+
+      - name: Login to DockerHub
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9  # v3.7.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      # Idempotency observability: warn (don't block) if the :version tag
+      # already exists on DockerHub. Re-push is safe because ECR is the source
+      # of truth and release tags are immutable there. We log the overwrite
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-dockerhub.yaml
```diff
@@ -0,0 +1,162 @@
...
+        run: |
+          ECR_IMAGE="${{ steps.login-ecr.outputs.registry }}/${{ github.event.repository.name }}:${{ inputs.releaseVersion }}"
+          docker pull "$ECR_IMAGE"
+          echo "ECR_IMAGE=$ECR_IMAGE" >> $GITHUB_ENV
+
+      - name: Login to DockerHub
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9  # v3.7.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
clarkbains: Repo has been onboarded so you can do use the following
```
   env:
        VAULT_ADDR: https://vault.maas-vault-prod.solace.cloud:8200
    steps:
      - name: Retrieve secrets from Vault
        id: secrets
        uses: hashicorp/vault-action@4c06c5ccf5c0761b6029f56cfb1dcf5565918a3b # v3.4.0
        continue-on-error: true
        with:
          url: "${{ env.VAULT_ADDR }}"
          role: "cicd-workflows-secret-read-role"
          method: jwt
          path: jwt-github
          jwtGithubAudience: https://github.com/${{ github.repository_owner }}
          exportToken: true
          secrets: |
            secret/data/tools/githubactions DOCKERHUB_USER | DOCKERHUB_USER ;
            secret/data/tools/githubactions DOCKERHUB_PASS | DOCKERHUB_PASS ;

      - name: Login to Docker Hub
        uses: docker/login-action@b45d80f862d83dbcd57f89517bcf500b2ab88fb2 # v4.0.0
        with:
          username: ${{ steps.secrets.outputs.DOCKERHUB_USER }}
          password: ${{ steps.secrets.outputs.DOCKERHUB_PASS }}
```

.github/workflows/publish-to-dockerhub.yaml
```diff
@@ -0,0 +1,162 @@
...
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@2f9f10ea3fa2eed41ac443fee8bfbd059af2d0a4  # v1.6.0
+
+      # Pre-flight: surface "ECR tag missing" as a human-readable error
+      # before we attempt the docker pull (whose 'manifest unknown' failure
+      # is opaque). The most likely cause of a missing tag is that the
+      # Release workflow's publish job hasn't run yet for this version.
+      - name: Verify ECR tag exists
+        env:
+          REGION: ${{ secrets.EMA_AWS_DEFAULT_REGION }}
```
clarkbains, line 65: Please follow this pattern everywhere as opposed to direct interpolation of inputs in `run` blocks

.github/workflows/publish-to-dockerhub.yaml
```diff
@@ -0,0 +1,162 @@
...
+            echo "::error::Refusing to push :latest for pre-release version '$VERSION'. Stable releases have no suffix."
+            exit 1
+          fi
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@5fd3084fc36e372ff1fff382a39b10d03659f355  # v2.2.0
+        with:
+          aws-access-key-id: ${{ secrets.EMA_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.EMA_AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.EMA_AWS_DEFAULT_REGION }}
```
clarkbains, line 54: What are these credentials?
Engplat-Infra maintains a standard way of granting programatic access to ECR repos without credentials, recommend connecting with them for how to do this.  (ECR Framework). The existing ECR repo already exists so it needs to be imported into state for this to work.

bol444: Good catch! These are long-lived static AWS keys (EMA_AWS_ACCESS_KEY_ID / _SECRET_ACCESS_KEY). They're not new tho — this workflow inherited them from the existing pattern, they're the same secrets used by deploy-managed-ema-image.yaml, deploy-dev-managed-ema-image.yaml, and release.yaml. 
So this PR isn't introducing a new credential.
I'm fully on board with moving to the credential-less ECR Framework, I am thinking we can approach this differently and plz let me know which way you prefer:
(a) We can adopt the ECR Framework in this workflow only, for now, and it leaves the workflow a bit inconsistent with the other three (they'd still use static keys)
(b) Keep `EMA_AWS_*` here to match the existing three workflows, and file a separate ticket to migrate all four to the ECR Framework in one pass (including the state import).
Either way, can you point me at the ECR Framework onboarding doc / the "import existing repo into state" steps so I can scope the follow-up ticket? @clarkbains

.github/workflows/publish-to-dockerhub.yaml
bol444: Done. I moved every input off direct `run:`-block interpolation and into `env:` bindings across all six run steps.

.github/workflows/publish-to-dockerhub.yaml
bol444: Done. I switched DockerHub auth to Vault. Now the creds come straight from `secret/data/tools/githubactions`.

A few thing I would like to flag here:
1. I added permissions: `id-token: write` — the JWT auth needs the GitHub OIDC token. (`release.yaml` already uses this for Guardian/FOSSA, so it's consistent with the repo.)
2. I dropped `continue-on-error: true` from the Vault step. Since the DockerHub creds are hard-required, I'd rather fail loudly at the Vault step than later when requested access at docker push three steps later. Happy to restore it if there's a reason it was in the template I'm missing.
3. I also bumped docker/login-action to the v4.0.0 SHA from your snippet (was v3.7.0).

One thing I want to confirm before this is truly done: does the `DOCKERHUB_USER` in that Vault path have push rights to `solace/event-management-agent` (not just pull)? @clarkbains

bol444:
@clarkbains The OIDC onboarding looks half-applied on event-management-agent, the`build.yaml`'s "Run Tests and Deploy" and "FOSSA Scan" failed with [OICD errors](https://github.com/SolaceProducts/event-management-agent/actions/runs/29430368396)

.github/workflows/publish-to-dockerhub.yaml
clarkbains: Yes, this token has write access

.github/workflows/publish-to-dockerhub.yaml
clarkbains: https://sol-jira.atlassian.net/wiki/spaces/HCI/pages/3526199503/ECR+repository+Framework#Allow-Workflows-to-push-to-the-ECR-repository
The document can be found above, although due to the need to import the existing repo it is not a self serve process. If the secrets are not new I do not think this has to be done as a part of this ticket

clarkbains:
@bol444 Separate issue, should be good to go now

.github/workflows/publish-to-dockerhub.yaml
bol444: Thx for the link @clarkbains , I created a follow up ticket to migration to the ECR Framework: https://sol-jira.atlassian.net/browse/DATAGO-145110.

bol444:
> @bol444 Separate issue, should be good to go now

If there are no other issues, could you approve the PR @clarkbains ?

CameronRushton (approved):

PR #313, opened 2026-06-25
DATAGO-139741: auto-derive developmentVersion in release workflow

bol444:
## What are the changes

- **Workflow fix:** remove the `developmentVersion` `workflow_dispatch` input and let `maven-release-plugin` auto-derive it as `<patch+1 of releaseVersion>-SNAPSHOT`. 
- Add a `Validate release version` step that rejects malformed inputs and surfaces the auto-derived value.
- **Cleanup commit:** bump main's 9 module poms from `1.9.9-SNAPSHOT` to `1.9.10-SNAPSHOT`, completing the version-bump that should have happened automatically during the v1.9.9 release.

Jira link: [DATAGO-139741](https://sol-jira.atlassian.net/browse/DATAGO-139741).

## Why the workflow fix

The v1.9.9 release ended up with `main`'s pom at `1.9.9-SNAPSHOT` instead of `1.9.10-SNAPSHOT` because the operator entered `developmentVersion=1.9.9-SNAPSHOT` (the snapshot form of the just-released version) instead of `1.9.10-SNAPSHOT` (the next-version snapshot). The workflow passed the value verbatim to `mvn`, which dutifully round-tripped the pom version.
The workflow fix removes this manual entering of `developmentVersion` value by the operator and let `maven-release-plugin` auto-derive it as `<patch+1 of releaseVersion>-SNAPSHOT`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,13 +3,9 @@ on:
   workflow_dispatch:
     inputs:
       releaseVersion:
-        description: "Default version to use when preparing a release."
+        description: "Version to release, in X.Y.Z form (no -SNAPSHOT, no placeholder). The next development version on main is auto-derived as X.Y.(Z+1)-SNAPSHOT by maven-release-plugin."
         required: true
         default: "A.B.C"
-      developmentVersion:
-        description: "Default version to use for new local working copy (the next version after version A.B.C)."
-        required: true
-        default: "X.Y.Z-SNAPSHOT"
       dry_run:
         description: "Dry run: run the readiness check end-to-end, skip the actual mvn release commit/push and the entire publish job (no GitHub Release, no ECR retag, no manifest DB write). Use for validating the workflow without altering the repo or registries."
         type: boolean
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yaml
```diff
@@ -3,13 +3,9 @@ on:
...
     inputs:
       releaseVersion:
-        description: "Default version to use when preparing a release."
+        description: "Version to release, in X.Y.Z form (no -SNAPSHOT, no placeholder). The next development version on main is auto-derived as X.Y.(Z+1)-SNAPSHOT by maven-release-plugin."
         required: true
         default: "A.B.C"
-      developmentVersion:
-        description: "Default version to use for new local working copy (the next version after version A.B.C)."
-        required: true
-        default: "X.Y.Z-SNAPSHOT"
```
bol444, line 12: Image

Removing developmentVersion (highlighted in the screenshot) and letting maven-release-plugin auto-derive the next-snapshot value as <patch+1>-SNAPSHOT (e.g., 1.9.11-SNAPSHOT when releasing 1.9.10). This eliminates the operator-error class that caused on the v1.9.9 release: the operator entered 1.9.9-SNAPSHOT in this field instead of 1.9.10-SNAPSHOT, and the release round-tripped

.github/workflows/release.yaml
```diff
@@ -56,6 +52,21 @@ jobs:
           java-version: 17
           distribution: 'temurin'
           cache: 'maven'
+      - name: Validate release version
```
bol444, line 55: Add a Validate release version step that rejects malformed inputs and surfaces the auto-derived value

.github/workflows/release.yaml
```diff
@@ -84,17 +95,16 @@ jobs:
...
         run: |
           echo "::warning::DRY RUN — Maven release skipped. Would have executed:"
           echo "  mvn release:prepare release:perform -B --file service/pom.xml \\"
-          echo "    -DreleaseVersion=${{ github.event.inputs.releaseVersion }} \\"
-          echo "    -DdevelopmentVersion=${{ github.event.inputs.developmentVersion }}"
+          echo "    -DreleaseVersion=${{ github.event.inputs.releaseVersion }}"
           echo ""
           echo "Side effects that would have occurred:"
-          echo "  - Commit pom.xml version bumps to ${{ github.event.inputs.releaseVersion }} (release) and ${{ github.event.inputs.developmentVersion }} (next snapshot)"
+          echo "  - Commit pom.xml version bumps to ${{ github.event.inputs.releaseVersion }} (release) and $DERIVED_DEVELOPMENT_VERSION (next snapshot, auto-derived)"
```
bol444, line 107: Surface the auto-derived next-snapshot value in the dry-run summary so the operator can sanity-check before a real run

CameronRushton (approved):

PR #326, opened 2026-09-10
DATAGO-151929 upgrade to java 25

chuanbao-solace:

    Upgraded to java 25 etc.

    ...

    ...

    ...


service/application/src/test/java/com/solace/maas/ep/event/management/agent/commandManager/CommandManagerTests.java
```diff
@@ -98,7 +98,7 @@ class CommandManagerTests {
     private EventPortalProperties eventPortalProperties;
 
     private static final String MESSAGING_SERVICE_ID = "myMessagingServiceId";
-    private final static ThreadPoolTaskExecutor testThreadPool = new ThreadPoolTaskExecutor();
+    private static final ThreadPoolTaskExecutor testThreadPool = new ThreadPoolTaskExecutor();
```
bol444, line 101: if this field is not used, as stated by SonarQube, let's just remove it

service/application/resources/pmd/pmd-rules.xml
```diff
@@ -4,6 +4,10 @@
         Event management agent custom rule set.
     </description>
 
+    <!-- PMD 7 (maven-pmd-plugin 3.28.0): rules renamed from their PMD 6 names, and
+         rules newly introduced by PMD 7 excluded to keep the pre-upgrade lint baseline.
+         Revisit the newly-excluded rules separately from the Java 25 upgrade. -->
```
bol444, line 9: suggest to update this comment to be more accurate:
The comment says the excludes are "rules renamed from their PMD 6 names, and rules newly introduced by PMD 7,", but a chunk of the newly-added excludes are actually **long-standing PMD 6 rules**, PMD 7's reworked type-resolution just makes them fire where PMD 6 didn't.

I see 3 categories for this PMD rules update:

a) New in PMD 7 — keep excluded (deferral is fine): LambdaCanBeMethodReference (7.1), AvoidSynchronizedStatement (7.5), ImplicitFunctionalInterface (7.12), UnnecessaryWarningSuppression (7.14), RelianceOnDefaultCharset (7.17), TypeParameterNamingConventions (7.17, replaces GenericsNaming), VariableCanBeInlined (replaces UnnecessaryLocalBeforeReturn), UselessPureMethodCall.

b) Renames carried over — no action: NonExhaustiveSwitch, UnitTest*, TooFewBranchesForSwitch.

c) **Pre-existing PMD 6 rules newly silenced** — these aren't renames or new-in-7: LooseCoupling (0.7), BooleanGetMethodName (4.0), UnnecessaryFullyQualifiedName (5.0), SimplifyBooleanReturns (0.9), CouplingBetweenObjects (1.04), ExcessivePublicCount (1.04), ExceptionAsFlowControl (1.8), ImmutableField (2.0), AvoidThrowingNewInstanceOfSameException (4.2.5), DataClass (6.0.0) — I think **these are fine to keep excluded** (as they mostly style/design opinion).  But there are 4 I'd push back on (see my next comment below). 

@chuanbao-solace Could we reword the header comment to name this third bucket so it's clear we're deferring pre-existing checks, not just new rules?

service/application/resources/pmd/pmd-rules.xml
```diff
@@ -4,6 +4,10 @@
...
+
```
bol444, line 10: These 4 excludes are pre-existing bug-catchers, not new PMD 7 rules, I recommend we keep them enforced rather than exclude:
- CloseResource — resource leaks
- InvalidLogMessageFormat — mismatched {} placeholders / broken log lines
- ConstructorCallsOverridableMethod — real init-order bug pattern
- AssignmentInOperand — flags legit idioms like while ((line = r.readLine()) != null))

I suggested we re-enable these 4 and see what we get and try to fix in code if we get PMD warnings on them.

service/application/src/main/java/com/solace/maas/ep/event/management/agent/subscriber/SolaceDirectMessageHandler.java
```diff
@@ -21,7 +21,7 @@ protected SolaceDirectMessageHandler(String topicString, SolaceSubscriber solace
         solaceSubscriber.registerMessageHandler(this);
     }
 
-    public String getTopicString() {
+    public final String getTopicString() {
```
mzhang444, line 24: What is the reason behind making this public method final? Doesn't it mean any public method defined in abstract class must be final to avoid being overridden by subclass?

chuanbao-solace: SolaceDirectMessageHandler's constructor (line 21) calls solaceSubscriber.registerMessageHandler(this), and SolaceSubscriber.registerMessageHandler immediately calls back solaceMessageHandler.getTopicString() (SolaceSubscriber.java:24,30) — while the constructor is still running.

If getTopicString() weren't final, PMD's ConstructorCallsOverridableMethod rule would flag this as a real bug pattern: a subclass could override getTopicString(), and since the base constructor invokes it via
  registerMessageHandler(this) before the subclass's own constructor body/field initializers have run, the override could read uninitialized subclass state (or return null/garbage). Making it final closes that hole — no subclass can override it, so there's no risk of the callback reaching un-initialized state during construction.

service/plugin/src/main/java/com/solace/maas/ep/event/management/agent/plugin/command/model/JobStatus.java
```diff
@@ -1,11 +1,8 @@
...
-    error,
-    validation_error,
-    success;
-
-    JobStatus() {
-    }
+    IN_PROGRESS,
+    ERROR,
+    VALIDATION_ERROR,
+    SUCCESS
```
bol444: any reason we are making these changes? as this will break config push.
EMA now sends "SUCCESS" to EP, but EP decodes command responses by enum name (case-sensitive) into its still-lowercase JobStatus, so the value silently becomes `null` and `ConfigPushProcessor` marks every config-push as validation_error.

bol444 (approved):

